PR #48, opened 2015-02-25
Update uthash.h

justanotheranonymoususer:
Removed a redundant semicolon (It is redundant, isn't it?).


troydhanson:
Good catch, thanks.

PR #50, opened 2015-04-06
Lint 20150403

fperrad:
this set of patches solves problems found with a lint tool (with some MISRA C 2012 rules).

I try to have a commit per kind of problem.

problems found in the test suite are in another pull request.


troydhanson:
Thanks, merged. Just curious, is there a reason for tests like tmp to be written tmp != NULL or just style?

fperrad:
see http://www.gimpel-online.com/MsgRef.html#9036 from the tool maker

```
if (a != 0);
if (ptr != NULL); 
```
is more readable (more explicit), so more safe than the idiom
```
if (a);
if (ptr);
```

PR #54, opened 2015-04-14
Lint tests 20150414

fperrad:


troydhanson:
I agree with 90% of these changes but I prefer statements like "if (verbose) ..." rather than turning those kind of statements into arithmetical tests. I am merging it anyway, since that is a bit of a nitpick.

fperrad:
since C99, the type `bool` was backported from C++, and it is defined in `stdbool.h`.

```
int a = 1;
bool verbose = true;
if (a); // bad
if (a != 0); // good
if (verbose); // good too
```

PR #56, opened 2015-04-18
Lint 20140417

fperrad:


troydhanson:
Thanks for the thorough tidying up

PR #57, opened 2015-04-20
Lint 20150420

fperrad:


zhangjikai:
If I can use long as the key of the uthash

troydhanson:
> On May 5, 2015, at 4:17 AM, zhangjikai <notifications@github.com> wrote:
> 
> If I can use long as the key of the uthash
> 
> 

Yes you can use a long as a key in uthash

PR #60, opened 2015-05-06
Fix locks

tbeu:
* In thread_routine_r of test1.c move unlock after access of e, such that another thread with wrlock does not delete/replace it if HASH_DEL is commented in again.
 * Destroy lock after successful use


troydhanson:
thanks!

PR #61, opened 2015-06-26
Add utringbuffer.h, a statically sized ring-buffer implementation

Quuxplusone:
Documentation: http://quuxplusone.github.io/uthash/utringbuffer.html

The need for a ring-buffer data structure came up at work, and we already use utarray, so I put together a quick copy-and-paste to create "utringbuffer". I don't know if uthash is in the business of collecting random data structures, but we do think it would be pretty great if this one made it into the "official" distribution.

utringbuffer models a "history" of the last `n` elements to have been pushed-back onto the buffer (where `n` is a parameter to `utringbuffer_new`). Once `utringbuffer_len(a)` reaches `n` (that is, once `utringbuffer_full(a)` starts returning `true`), each new pushed-back element *destroys and overwrites* the current oldest element. Iterating from `utringbuffer_front(a)` to `utringbuffer_back(a)` iterates from oldest to newest; i.e., push-back pushes onto the back of the queue just the way its name suggests. There is no way to "pop" elements; once the buffer becomes full, the only way to make it un-full is to call `utringbuffer_clear(a)`.

The copyright header on `utringbuffer.h` already says "Troy D. Hansen" (as opposed to "Arthur O'Dwyer") merely to minimize the work of merging into this repo; I'm happy to give away my rights to the code, given that it would still be BSD-licensed.

### DIFF
--- a/doc/index.html
+++ b/doc/index.html
@@ -29,6 +29,7 @@
         <div><a href="userguide.html">uthash</a></div>
         <div><a href="utlist.html">utlist</a></div>
         <div><a href="utarray.html">utarray</a></div>
+        <div><a href="utringbuffer.html">utringbuffer</a></div>
         <div><a href="utstring.html">utstring</a></div>
 
         <h2>activity</h2>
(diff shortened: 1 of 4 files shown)


troydhanson:
Nice work Arthur. It seems handy to include it with the official ut headers. I would like to add some unit tests. If you get a chance and want to add to your pull request feel free otherwise I can whip up a couple. That keeps me from breaking the code in the future. 

Thanks for documenting it! 

It may take a week or two before I'm in a position to actually merge it.

PR #64, opened 2015-11-03
Fix realloc and oom

tbeu:
With these changes it is now possible to have `#define oom() break` and do not have memory leaks in case `realloc` fails.

### DIFF
--- a/src/utarray.h
+++ b/src/utarray.h
@@ -38,7 +38,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <string.h>  /* memset, etc */
 #include <stdlib.h>  /* exit */
 
+#ifndef oom
 #define oom() exit(-1)
+#endif
 
 typedef void (ctor_f)(void *dst, const void *src);
 typedef void (dtor_f)(void *elt);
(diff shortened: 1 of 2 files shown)


tbeu:
There still is an issue with name clashes of the tmp buffer of `utstring_reserve` and the UT_string, i.e., 
```c
#include "utstring.h"
UT_string utstring_tmp;
utstring_init(&utstring_tmp);
```
gives compiler errors.

troydhanson:
I think the trouble with this approach is that if oom is defined as break, then the reserve operation silently fails. The subsequent macros expect reserve to have always succeeded (or the program aborts).

tbeu:
I thougt the reason why you have those `do .. while(0)` dummy loops is to allow `oom` and `hash_fata`l to be defined as `break`.

troydhanson:
The do-while loops are there to make the macro work as a statement , see for example http://stackoverflow.com/questions/1067226/c-multi-line-macro-do-while0-vs-scope-block.

I think its only safe to redefine the oom macro to abort or something like a siglongjmp out of scope. Using break there is almost certainly going to cause a program crash. (Of course OOM means other troubles anyway).

tbeu:
I need the `break` there to clean up locked resources before doing the siglongjmp out of scope.

PR #66, opened 2015-11-26
fix, need free the element which is from malloc

cheyiliu:


Quuxplusone:
Looks to me like this was merged in 044474afe5a57db4302cc722de476d20d1ab70c7 (and the docs regenerated in 1d53c03a).

PR #67, opened 2016-02-22
API compatible extensions to functionality of UTList macros

thiloschulz:
I have recently updated my project to use utlist macros for list management, because I found the concept to be pretty neat, however, there was some functionality missing for me.

In the three commits here presented (ordered by decreasing likelihood of you pulling them :)
- *_APPEND_ELEM* class of functions in addition to preexisting *_PREPEND_ELEM*
- Added a few *2 functions for user defined list pointer fields that were missing from some functions such as *_REPLACE_ELEM2
- Allow elt to be a NULL pointer for the _PREPEND_ELEM (and the new _APPEND_ELEM) macros. In this case the item will be appended (prepended) to the list.

Best regards,
Thilo Schulz

### DIFF
--- a/doc/utlist.html
+++ b/doc/utlist.html
@@ -3,7 +3,7 @@
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en">
 <head>
 <meta http-equiv="Content-Type" content="application/xhtml+xml; charset=UTF-8" />
-<meta name="generator" content="AsciiDoc 8.6.7" />
+<meta name="generator" content="AsciiDoc 8.6.9" />
 <title>utlist: linked list macros for C structures</title>
 <style type="text/css">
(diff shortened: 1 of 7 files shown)


troydhanson:
Hi Thilo, I took a very quick look, it seems reasonable. What I would need is an update to the documentation and unit tests for the new macros and changes to the existing ones. It's the only way to detect breakage from future changes and validate expected behavior. If you have time please feel free to include otherwise I can take a stab at it. It may be a while before I can review fully and merge. Thanks.

thiloschulz:
I would be more than happy to do this. Thank you for your consideration and for the heads-up, I really wanted to see your reaction first before doing anything further.

I take it that the .html is auto-generated from the respective .txt file?

thiloschulz:
Alright, please check out my latest commits.

troydhanson:
Thanks Thilo. I've merged it.

PR #71, opened 2016-06-13
Support adding elements to hash in ordered fashion

thiloschulz:
At present, the linked list maintained by the HASH_ADD* function family will only be sorted upon invocation of the HASH_SRT() macro.
I have extended the HASH_ADD* macro family by introduction of HASH_SADD* macros which have an argument cmp for the compare function, allowing proper sorting upon adding elements to the hash.

Fully backwards compatible.

Please tell me what you think, if you give the okay I can write a testcase and update the documentation.

Best regards,
Thilo

### DIFF
--- a/doc/userguide.txt
+++ b/doc/userguide.txt
@@ -370,7 +370,7 @@ The hash table pointer (which initially points to the first item added to the
 hash) can change in response to `HASH_DEL` (i.e. if you delete the first item
 in the hash table).
 
-Iteractive deletion
+Iterative deletion
 ^^^^^^^^^^^^^^^^^^
 The `HASH_ITER` macro is a deletion-safe iteration construct which expands
 to a simple 'for' loop.
--- a/doc/userguide.txt
+++ b/doc/userguide.txt
@@ -370,7 +370,7 @@ The hash table pointer (which initially points to the first item added to the
 hash) can change in response to `HASH_DEL` (i.e. if you delete the first item
 in the hash table).
 
-Iterative deletion
+Iteractive deletion
 ^^^^^^^^^^^^^^^^^^
 The `HASH_ITER` macro is a deletion-safe iteration construct which expands
 to a simple 'for' loop.
@@ -1141,6 +1141,33 @@ always used with the `users_by_name` hash table).
 ----------------------------------------------------------------------
 
 
+Sorted insertion of new items
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you would like to maintain a sorted hash you have two options. The first
+option is to use the HASH_SRT() macro, which will sort any unordered list in
+'O(n log(n))'. This is the best strategy if you're just filling up a hash
+table with items in random order with a single final HASH_SRT() operation
+when all is done. Obviously, this won't do what you want if you need
+the list to be in an ordered state at times between insertion of
+items. You can use HASH_SRT() after every insertion operation, but that will
+yield a computational complexity of 'O(n^2 log(n)'.
+
+The second route you can take is via the sorted add macros.
+The HASH_SADD*() macros work just like their HASH_ADD*() counterparts, but
+with an additional compare function argument:
+
+  int name_sort(struct my_struct *a, struct my_struct *b) {
+    return strcmp(a->name,b->name);
+  }
+
+  HASH_SADD_KEYPTR(hh, items, &item->name, strlen(item->name), item, name_sort);
+
+New items are sorted at insertion time in 'O(n)', thus resulting in a
+total computational complexity of 'O(n^2)' for the creation of the hash
+table with all items.
+For sorted add to work, the list must be in an ordered state before
+insertion of the new item.
+
 Several sort orders
 ~~~~~~~~~~~~~~~~~~~
 It comes as no surprise that two hash tables can have different sort orders, but
@@ -1500,7 +1527,7 @@ ineffective at reducing the chain lengths.
 Imagine a very bad hash function which always puts every item in bucket 0. No
 matter how many times the number of buckets is doubled, the chain length of
 bucket 0 stays the same. In a situation like this, the best behavior is to 
-stop expanding, and accept O(n) lookup performance. This is what uthash
+stop expanding, and accept 'O(n)' lookup performance. This is what uthash
 does. It degrades gracefully if the hash function is ill-suited to the keys.
 
 If two consecutive bucket expansions yield `ideal%` values below 50%, uthash
@@ -1712,18 +1739,20 @@ than `hh`, or if your key's data type isn't `int` or `char[]`.
 .General macros
 [width="90%",cols="10m,30m",grid="none",options="header"]
 |===============================================================================
-|macro          | arguments
-|HASH_ADD       | (hh_name, head, keyfield_name, key_len, item_ptr)
-|HASH_ADD_KEYPTR| (hh_name, head, key_ptr, key_len, item_ptr)
-|HASH_REPLACE   | (hh_name, head, keyfield_name, key_len, item_ptr, replaced_item_ptr)
-|HASH_FIND      | (hh_name, head, key_ptr, key_len, item_ptr)
-|HASH_DELETE    | (hh_name, head, item_ptr)
-|HASH_SRT       | (hh_name, head, cmp)
-|HASH_CNT       | (hh_name, head)
-|HASH_CLEAR     | (hh_name, head)
-|HASH_SELECT    | (dst_hh_name, dst_head, src_hh_name, src_head, condition)
-|HASH_ITER      | (hh_name, head, item_ptr, tmp_item_ptr)
-|HASH_OVERHEAD  | (hh_name, head)
+|macro            | arguments
+|HASH_ADD         | (hh_name, head, keyfield_name, key_len, item_ptr)
+|HASH_ADD_KEYPTR  | (hh_name, head, key_ptr, key_len, item_ptr)
+|HASH_SADD        | (hh_name, head, keyfield_name, key_len, item_ptr, cmp)
+|HASH_SADD_KEYPTR | (hh_name, head, key_ptr, key_len, item_ptr, cmp)
+|HASH_REPLACE     | (hh_name, head, keyfield_name, key_len, item_ptr, replaced_item_ptr)
+|HASH_FIND        | (hh_name, head, key_ptr, key_len, item_ptr)
+|HASH_DELETE      | (hh_name, head, item_ptr)
+|HASH_SRT         | (hh_name, head, cmp)
+|HASH_CNT         | (hh_name, head)
+|HASH_CLEAR       | (hh_name, head)
+|HASH_SELECT      | (dst_hh_name, dst_head, src_hh_name, src_head, condition)
+|HASH_ITER        | (hh_name, head, item_ptr, tmp_item_ptr)
+|HASH_OVERHEAD    | (hh_name, head)
 |===============================================================================
 
 [NOTE]
--- a/src/uthash.h
+++ b/src/uthash.h
@@ -171,11 +171,8 @@ do {                                                                            
   (head)->hh.tbl->signature = HASH_SIGNATURE;                                    \
 } while(0)
 
-#define HASH_ADD(hh,head,fieldname,keylen_in,add)                                \
-    HASH_SADD(hh,(head),fieldname,(keylen_in),(add),NULL)
- 
 #define HASH_SADD(hh,head,fieldname,keylen_in,add,cmpfcn)                        \
-    HASH_SADD_KEYPTR(hh,(head),&((add)->fieldname),(keylen_in),(add),(cmpfcn))
+    HASH_SADD_KEYPTR(hh,(head),&((add)->fieldname),(keylen_in),(add),cmpfcn)
 
 #define HASH_REPLACE(hh,head,fieldname,keylen_in,add,replaced)                   \
 do {                                                                             \
@@ -187,9 +184,6 @@ do {                                                                            
   HASH_ADD(hh,(head),fieldname,(keylen_in),(add));                               \
 } while(0)
 
-#define HASH_ADD_KEYPTR(hh,head,keyptr,keylen_in,add)                            \
-    HASH_SADD_KEYPTR(hh,(head),(keyptr),(keylen_in),(add),NULL)
-
 #define HASH_APPEND_LIST(hh, head, add)                                          \
 do {                                                                             \
   (add)->hh.next = NULL;                                                         \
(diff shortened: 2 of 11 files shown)


troydhanson:
Hi Thilo I've been away a while. Is the SADD the same as an ADD followed by an immediate SRT?

thiloschulz:
When there is a requirement that the hash table's list is sorted at all times, i.e. immediately after every insertion of a new entry, this would scale O(n^2 log(n)) as there'd be a new merge sort for every new item that is inserted.
So nope. My changes take care of the sorting at the same time as the insertion into the hash table, which scales O(n^2).

troydhanson:
I see, ok. Have you tested it on both gcc and Visual Studio? I often had to struggle with getting clean compiles across the two platforms. I only took a quick look, I can tell you spent a lot of time on it and included the unit test. I think you could go ahead with the documentation (I didn't see it in my quick look).

thiloschulz:
Thank you. I could not test it in Visual Studio, as I don't have Windows. There's nothing too wild going on there, though. Will write the doc now.

troydhanson:
Ok, please take your time. It may be next week by the time I can review the patch.

tbeu:
I can test it with VS.

thiloschulz:
Done. Also check out the .gitignore that is now autogenerated in the tests directory via Makefile which will keep the test output from clogging the git status report.

troydhanson:
First, thanks for your patience. And, thanks @tbeu for testing with VS. I have a small problem: I merged #72 first and now there are merge conflicts. I know this is asking a lot but if you have a chance, and could update your patch against the current head, it would help me out a lot.

thiloschulz:
oh well, this is a mess. I'll do my best.

thiloschulz:
I have done as you asked.
I've removed some redundancy in the *ADD_KEYPTR and *ADD_KEYPTR_BY_HASH_VALUE,
also I abbreviated *BY_HASH_VALUE to *BY_HVAL

Given that you merged less well tested, more extensive API changing code, that was written long **after** my pull request, I'd appreciate if you compelled the author of pull request #72 to extend my unit test87 to test the various new SADD_*BY_HVAL macros, and fix any problems that might arise.

src/uthash.h
```diff
@@ -188,79 +190,117 @@ do {
   (head)->hh.tbl->signature = HASH_SIGNATURE;                                    \
 } while(0)
 
-#define HASH_ADD(hh,head,fieldname,keylen_in,add)                                \
-        HASH_ADD_KEYPTR(hh,head,&((add)->fieldname),keylen_in,add)
+#define HASH_SADD(hh,head,fieldname,keylen_in,add,cmpfcn)                        \
+    HASH_SADD_KEYPTR(hh,(head),&((add)->fieldname),(keylen_in),(add),cmpfcn)
+
+#define HASH_SADD_BYHVAL(hh,head,fieldname,keylen_in,hashval,add,cmpfcn)         \
```
Quuxplusone: This should read `..._BY_HVAL` under your new naming convention; but why did you change it at all? Is something wrong with `..._BY_HASH_VALUE`?

src/uthash.h
```diff
@@ -188,79 +190,117 @@ do {
...
+  HASH_TO_BKT((hashval), (head)->hh.tbl->num_buckets, _ha_bkt);                  \
+  HASH_ADD_TO_BKT(head->hh.tbl->buckets[_ha_bkt],&add->hh);                      \
+  HASH_BLOOM_ADD(head->hh.tbl,add->hh.hashv);                                    \
+  HASH_EMIT_KEY(hh,head,keyptr,keylen_in);                                       \
+  HASH_FSCK(hh,head);                                                            \
 } while(0)
 
+#define GHV_DIRECT(hh, add, hashval)                                             \
+  (add)->hh.hashv = (hashval)
+#define GHV_COMPUTE(hh, keyptr, keylen_in, add)                                  \
```
Quuxplusone: I believe both of these macros should start with `HASH_` for namespacing/scoping reasons. What do they actually do, btw? What would go wrong if we didn't have these macros?

tests/test32.c
```diff
@@ -10,10 +10,10 @@ typedef struct el {
     struct el *next, *prev;
 } el;
 
-static int namecmp(el *a, el *b)
+/* static int namecmp(el *a, el *b)
 {
     return strcmp(a->bname,b->bname);
-}
+} */
```
Quuxplusone: If this is dead code, could you just remove it instead of commenting it out?

tests/test34.c
```diff
@@ -10,10 +10,10 @@ typedef struct el {
     struct el *next, *prev;
 } el;
 
-static int namecmp(el *a, el *b)
+/* static int namecmp(el *a, el *b)
 {
     return strcmp(a->bname,b->bname);
-}
+} */
```
Quuxplusone: Remove dead code instead of commenting it out?

tests/test65.c
```diff
@@ -28,6 +29,7 @@ static char * /*value*/ find_in_cache(const char *key)
     }
     return NULL;
 }
+#endif
```
Quuxplusone: Remove dead code instead of commenting it out?

src/uthash.h
thiloschulz: unnecessarily long.

src/uthash.h
thiloschulz: They were part of an attempt to integrate your code into my changes and merge the ADD_KEYPTR and ADD_KEYPTR_BY_HASH_VALUE into one function.
We could go without these macros, but they don't hurt either

troydhanson:
All, I apologize for the mess. I have been too much disconnected from uthash to be able to properly study the merges. I have reverted the merge of #72. In fact, after the merge, I also question the value (no pun intended) of the hash value-based functions. I did some tests years ago that gave me no performance gain from a similar change, but if you think I'm wrong, say so. @thiloschulz I'm sorry for merging less well tested code prior to yours. I naively thought that I'd be able to resolve the conflicts easily only to find they weren't, and in any case did not take the time to evaluate it properly. Which leads to my other point- I list on the uthash site that it's unmaintained for a reason. I really don't have time to maintain it properly. I mostly want it to stay functional, which means not changing it- or finding co-maintainers. (Do I hear volunteers?) I found the use case for @thiloschulz's feature compelling (maintaining a sorted list), so to resolve the situation: my thinking right now, is to merge the original SADD pull request, only (skip #72). Do let me know your thoughts, and please forgive my clumsy handling of these pull requests.

Quuxplusone:
This might be better taken to email; mine is arthur.j.odwyer@gmail.com if you want to.

> I question the value of the hash value-based functions. I did some tests years ago that gave me no performance gain from a similar change, but if you think I'm wrong, say so.

Definitely wrong; my employer has recently adopted something like #72 / #75 for *massive* speed gains, in use-cases where we're repeatedly looking up ~64-character strings in two different hashes inside a relatively-tight-inner loop. Obviously there's no point in pulling out the `HASH_VALUE` computation if you're only going to `HASH_FIND_BY_HASH_VALUE` once; but if you're going to do it any more than once, there's no point in *not* pulling it out. Re-computing a Jenkins hash that you *have already computed once* is just silly.

> Which leads to my other point- I list on the uthash site that it's unmaintained for a reason. I really don't have time to maintain it properly. I mostly want it to stay functional, which means not changing it- or finding co-maintainers.

Agreed. I would love to see it changed less often than it is, and was surprised to see #72 merged without discussion. I'm happy to pitch in as a part-time maintainer/contributor, but lack the ability to test on Windows or various obscure platforms, which I understand are important demographics for uthash. I'm definitely of the opinion that "stagnant but working project" is much much better than "actively updated but perpetually broken project". I don't want to see uthash turn into https://github.com/botwillacceptanything/botwillacceptanything! :)

doc/userguide.txt
```diff
@@ -1141,6 +1141,33 @@ always used with the `users_by_name` hash table).
...
+Sorted insertion of new items
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you would like to maintain a sorted hash you have two options. The first
+option is to use the HASH_SRT() macro, which will sort any unordered list in
+'O(n log(n))'. This is the best strategy if you're just filling up a hash
+table with items in random order with a single final HASH_SRT() operation
+when all is done. Obviously, this won't do what you want if you need
+the list to be in an ordered state at times between insertion of
+items. You can use HASH_SRT() after every insertion operation, but that will
+yield a computational complexity of 'O(n^2 log(n)'.
```
Quuxplusone: `'O(n^2 log(n))'` (missing parenthesis)

src/uthash.h
```diff
@@ -188,79 +190,117 @@ do {
...
+  HASH_EMIT_KEY(hh,head,keyptr,keylen_in);                                       \
+  HASH_FSCK(hh,head);                                                            \
 } while(0)
 
+#define HASH_GHV_DIRECT(hh, add, hashval)                                        \
+  (add)->hh.hashv = (hashval)
+#define HASH_GHV_COMPUTE(hh, keyptr, keylen_in, add)                             \
+  HASH_FCN((keyptr), (keylen_in), (add)->hh.hashv)
+
+#define HASH_ADD_SORTED(hh,head,keyptr,keylen_in,hashval,add,cmpfcn)             \
```
Quuxplusone: `HASH_ADD_SORTED` is much more likely to be mistaken by the user for "a sorted version of `HASH_ADD`" than `HASH_SADD`. Could you not reverse the naming convention here so that the spelled-out version is the "public" one and the cryptic short version is the "private" one? This would be more in keeping with e.g. `HASH_FUNCTION` (public) versus `HASH_FCN` (private)...

src/uthash.h
```diff
@@ -188,79 +190,117 @@ do {
...
+    HASH_ADD_PROLOGUE(hh,(head),(keyptr),(keylen_in),(add))                      \
+    struct UT_hash_handle *iter = &(head)->hh;                                   \
```
Quuxplusone: This local variable needs to be named something like `_hs_iter` to avoid stomping on the client code's namespace; for example, they may have a macro named `iter` in their own code. (The convention in `HASH_FIND` is `_hf_foo` and the convention in `HASH_ADD` is `_ha_foo`, so I infer that the convention in `HASH_SADD` should be `_hs_foo`.)

src/uthash.h
```diff
@@ -188,79 +190,117 @@ do {
...
+        (head) = (add);                                                          \
+      iter->prev = add;                                                          \
+    }                                                                            \
+    else                                                                         \
+      HASH_ADD_EPILOGUE(hh,(head),(keyptr),(keylen_in),(hashval),(add))
+
+#define HASH_SADD_KEYPTR_BY_HVAL(hh,head,keyptr,keylen_in,hashval,add,cmpfcn)    \
+do { unsigned _ha_bkt;                                                           \
+    HASH_GHV_DIRECT(hh, add, hashval);                                           \
+    HASH_ADD_SORTED(hh,head,keyptr,keylen_in,hashval,add,cmpfcn)                 \
```
Quuxplusone: I think you're missing a `HASH_ADD_EPILOGUE` right here.
If this were my code, I'd say get rid of as many of these unsanitized "helper" macros as you can; C macros that contain unmatched braces are the devil.

src/uthash.h
```diff
@@ -188,79 +190,117 @@ do {
...
+
+#define HASH_SADD_KEYPTR(hh,head,keyptr,keylen_in,add,cmpfcn)                    \
+do { unsigned _ha_bkt;                                                           \
+    HASH_GHV_COMPUTE(hh, keyptr, keylen_in, add);                                \
+    HASH_ADD_SORTED(hh,head,keyptr,keylen_in,(add)->hh.hashv,add,cmpfcn)
```
Quuxplusone: Another missing `HASH_ADD_EPILOGUE`, I think

tests/Makefile
```diff
@@ -71,6 +71,9 @@ all: $(PROGS) $(UTILS) $(PLAT_UTILS) $(FUNCS) $(SPECIAL_FUNCS) $(TEST_TARGET)
 
 tests_only: $(PROGS) $(TEST_TARGET)
 
+GITIGN = .gitignore
+MKGITIGN = [ -f "$(GITIGN)" ] || echo "$(GITIGN)" > $(GITIGN); grep -q '^\$@$$' $(GITIGN) || echo "$@" >> $(GITIGN)
```
Quuxplusone: What does this do, and why?  It looks like it's echoing something into the end of (`>>`) the `.gitignore` file, but that doesn't make sense, because `.gitignore` is supposed to be version-controlled. `make test` shouldn't *modify* a version-controlled file, should it?

tests/Makefile
```diff
@@ -112,5 +118,5 @@ astyle:
 .PHONY: clean astyle
 
 clean:	
-	rm -f $(UTILS) $(PLAT_UTILS) $(PROGS) test*.out keystat.??? example *.exe
+	rm -f $(UTILS) $(PLAT_UTILS) $(PROGS) test*.out keystat.??? example *.exe $(GITIGN)
```
Quuxplusone: Likewise here: this looks like it's `rm`'ing the `.gitignore` file.

tests/test65.c
```diff
@@ -16,19 +16,6 @@ struct CacheEntry {
 };
 struct CacheEntry *cache = NULL;
 
-static char * /*value*/ find_in_cache(const char *key)
```
Quuxplusone: On closer inspection, this seems like useful example code. If you're trying to silence a compiler warning about an unused `static` function, how about just removing the `static` keyword, or adding a call to this function in `main`?

tests/test87.c
```diff
@@ -0,0 +1,75 @@
+#include <stdio.h>
+#include <uthash.h>
```
Quuxplusone: #include "uthash.h"

tests/test87.c
```diff
@@ -0,0 +1,75 @@
...
+  char name[32];
+  int weight;
+} hstruct_t;
+
+#define ARRAY_LEN(x) (sizeof(x) / sizeof(*(x)))
+hstruct_t tst[] = {{{}, "muh1", 2}, {{}, "muh2", 8}, {{}, "muh3", 1}, {{}, "muh4", 8}, {{}, "muh5", 3}, {{}, "muh6", 5},
+  {{}, "muh7", 6}, {{}, "muh8", 15}, {{}, "muh9", 6}, {{}, "muh10", 9}, {{}, "muh11", 10}, {{}, "muh12", 43}};
+hstruct_t *hTable;
+
+static int cmpfunc(hstruct_t *s1, hstruct_t *s2)
```
Quuxplusone: Technically, this should take `const hstruct_t *` parameters. It could be a one-liner if you wanted:

    static int cmpfunc(const hstruct_t *a, const hstruct_t *b)
    {
      return (a->weight < b->weight) ? -1 : (a->weight > b->weight);
    }

tests/test87.c
```diff
@@ -0,0 +1,75 @@
...
+  printtable();
+}
+
+int main()
+{
+  int index;
+
+  for(index = 0; index < ARRAY_LEN(tst); index++)
+  {
+    HASH_SADD(hh, hTable, name[0], strlen(tst[index].name), &tst[index], cmpfunc);
```
Quuxplusone: I get this and I think it's correct, but the use of `name[0]` is confusing. This is why uthash provides `HASH_ADD_STR`, right? Should you provide a `HASH_SADD_STR` convenience macro as well? (Of course I'd personally spell that `HASH_ADD_STR_SORTED` or some such.)

tests/test87.c
```diff
@@ -0,0 +1,75 @@
...
+  printf("###\n");
+}
+
+void delitem(const char *name)
+{
+  hstruct_t *item;
+  
+  HASH_FIND_STR(hTable, name, item);
+  HASH_DEL(hTable, item);
+  printtable();
```
Quuxplusone: This `printtable` after each delete seems like overkill. Personally, I'd just print the table once, after I'd inserted all the data; if it's correct then, then it'll have been correct after each add and it'll still be correct after each delete.

src/uthash.h
```diff
@@ -188,79 +190,117 @@ do {
...
+    HASH_ADD_PROLOGUE(hh,(head),(keyptr),(keylen_in),(add))                      \
+    struct UT_hash_handle *iter = &(head)->hh;                                   \
+    (add)->hh.tbl = (head)->hh.tbl;                                              \
+    do {                                                                         \
+      if(cmpfcn(DECLTYPE(head) ELMT_FROM_HH((head)->hh.tbl, iter), add) > 0)     \
+        break;                                                                   \
+    } while((iter = iter->next));                                                \
+    if(iter) {                                                                   \
+      (add)->hh.next = iter;                                                     \
+      if(((add)->hh.prev = iter->prev))                                          \
```
Quuxplusone: This is a null pointer dereference when `iter == NULL`, isn't it? I think this is a bug (and you should add a unit test exercising this codepath).

tests/Makefile
thiloschulz: All it does is autogenerate a .gitignore file automatically updated with all built executabled.

.gitignore may be version controlled, however, this is optional. There is a reason why the user must do "git add" on it before it's version controlled.

tests/Makefile
thiloschulz: make clean gets rid of autogenerated stuff.

tests/test87.c
thiloschulz: Yes it is a tad confusing. I don't like the convenience functions though. They make the implicit assumption that your hash handle is named "hh" inside your struct.

tests/test87.c
thiloschulz: Maybe so. I think a bit more verbosity doesn't hurt though. This is a unit test, not a twitter message. It will also help you better isolate where things go south in case of a bug.

src/uthash.h
thiloschulz: &head->hh is guaranteed to be valid after PROLOGUE. HASH_SADD does exercise this codepath.

src/uthash.h
thiloschulz: HASH_ADD_SORTED contains HASH_ADD_EPILOGUE.

src/uthash.h
thiloschulz: You're right, it's not terribly beautiful. On the other hand, you'd end up repeating the same code 6 times to get the same result.

src/uthash.h
thiloschulz: You're right.

src/uthash.h
thiloschulz: Sure.

src/uthash.h
Quuxplusone: Whoops, never mind; I had missed the `if(iter)` on line 266.

tests/test65.c
thiloschulz: I thought so too, that's why I left it in as a comment in case the later unit test should be extended. By only removing the static keyword you will not only have dead code in the .c file, but also in the resultant binary. Given your penchant for small diffs on the autogenerated docs, I'm a bit surprised you'd want to go this route ;)

Quuxplusone: > Given your penchant for small diffs on the autogenerated docs, I'm a bit surprised you'd want to go this route

Eh? Removing the `static` keyword is a single-line diff; and I'm not worried about the code size of the *unit test* binaries.

thiloschulz: Well, then, a bit more verbosity on those printtables isn't going to be a problem either :)

Quuxplusone: Well, now I think you're just being difficult. The lines-of-code size of the *checked-in code* for test87.ans *does* matter to me (and also it would be nice to focus that test on just the new feature being added, instead of having the vast majority of it devoted to testing `HASH_DEL`). The size of the *generated* code is what I'm saying doesn't matter.

thiloschulz: > The lines-of-code size of the checked-in code for test87.ans does matter to me

Seriously? We're not living in the age of 2M RAM and 14.4k dialup.
You're really concentrating on the wrong issue here.

You're right when you're saying there can be a more extensive test of the new macros. That unit test was from before your additions though, and was then sufficient.

thiloschulz:
> Obviously there's no point in pulling out the HASH_VALUE computation if you're only going to HASH_FIND_BY_HASH_VALUE once; but if you're going to do it any more than once, there's no point in not pulling it out. Re-computing a Jenkins hash that you have already computed once is just silly.

This is a problem if you have two disconnected structures in separate hashes but the same key.
I have to concur here, this is a valid point.

doc/userguide.txt
thiloschulz: thanks

tests/test87.c
thiloschulz: true.

troydhanson:
I'm sorry, I'm missing something: what is the issue about disconnected structures in separate hashes but the same key with regard to HASH_VALUE?

thiloschulz:
I meant to say that you will get a performance gain from quuxplusone's modifications if you have two separate structures, that are in two separate hash tables but have the same hash value. That is the use case where you will get a performance gain.

troydhanson:
I see, that makes sense. Ok, let me see if I understand the situation: if I pull your repo now, I'll get the integrated set of both changes- your SADD and Arthur's HVAL stuff. (I prefer that shorter name too.). The SADD feature is documented and has unit tests. But HVAL still needs docs and unit tests. Right, @Quuxplusone? We could layer those on as a separate commit? I'm trigger happy, and want to pull the changes while everyone is in agreement. Am I missing any downside of pulling @thiloschulz's repo now?

thiloschulz:
> if I pull your repo now, I'll get the integrated set of both changes- your SADD and Arthur's HVAL stuff.

Exactly.

My suggestion on how you could proceed: You will first need to pull Arthur's version again, then you can pull mine because the latest version of my repository is based off that one. the new HVAL macros do have some documentation, at least the macro functions are listed.
This would be a common starting point that quux can then base off his next pull request.

Unit test would still be required to be written.
Quuxplusone has made many suggestions which I have not implemented yet but that should be done. I can do them, or Quuxplusone can do them in his pull request #75.

troydhanson:
Ok! Merged in with 650214c (brings back memories of the Apple II CPU, the Motorola 6502). Thank you so much guys. I added a very minimal note about the HVAL macros in the userguide. I also fixed hashscan so it's compatible with the new macro prototypes. I also deleted the .html files from the repo.

PR #72, opened 2016-06-21
Add HASH_VALUE and HASH_FIND_BY_HASH_VALUE primitives; breaking change to HASH_FCN

Quuxplusone:
This patch separates the "hash a key into an `unsigned` hashvalue" step (now known as `HASH_VALUE`) from the "modulo a hashvalue into a bucket" step (still known as `HASH_TO_BKT`). This lets us remove two arguments from each of the `HASH_JEN` etc. hash function macros, because they no longer need to care about the number of buckets in the table.

The performance improvement here is that client code can compute the `HASH_VALUE` for a given key just once, store it locally, and then use that hashvalue to search for the same key in many different hash tables — or to search for it once, and then quickly insert it when it's not found — or to search for it multiple times in the same hash table.

Notice that the third commit in this PR is an API-breaking change for clients who are currently passing `-DHASH_FUNCTION=some_custom_function`. It is *not* a breaking change for clients passing `-DHASH_FUNCTION=HASH_SAX` or whatever; and it's not a quietly breaking change for anyone AFAICT.

### DIFF
--- a/doc/userguide.txt
+++ b/doc/userguide.txt
@@ -1712,18 +1712,23 @@ than `hh`, or if your key's data type isn't `int` or `char[]`.
 .General macros
 [width="90%",cols="10m,30m",grid="none",options="header"]
 |===============================================================================
-|macro          | arguments
-|HASH_ADD       | (hh_name, head, keyfield_name, key_len, item_ptr)
-|HASH_ADD_KEYPTR| (hh_name, head, key_ptr, key_len, item_ptr)
-|HASH_REPLACE   | (hh_name, head, keyfield_name, key_len, item_ptr, replaced_item_ptr)
-|HASH_FIND      | (hh_name, head, key_ptr, key_len, item_ptr)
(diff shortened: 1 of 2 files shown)


troydhanson:
I’m just getting back to the connected world. In principle it sounds fine. The breakage for custom function seems ok to me- never heard of anyone doing that. Hope to look at it in near future.

troydhanson:
It would be terrific if you could also add a unit test. It also serves as secondary purpose as a usage example.

thiloschulz:
Unit test hashscan fails to compile on latest master.

Quuxplusone:
@troydhanson I haven't looked at unit tests at all yet. Based on @thiloschulz's comment, it would IMHO be appropriate to revert until the test failure can be investigated/fixed.

thiloschulz:
Maybe not necessary, just an API breakage. Your job to investigate :-)

Quuxplusone:
Reopened as #75. Let's continue some of this discussion over there.

PR #73, opened 2016-06-26
Cast all integer user inputs.

jonathanmarvens:


troydhanson:
Hi Jonathan, I think I'd want another set of parens around the j so its like ((unsigned)(j)) instead of ((unsigned)j). Just in case j is a complex argument. Incidentally why do you need this? Is it to quiet a warning if j is some other sized type?

jonathanmarvens:
@troydhanson: I've pushed an updated commit that's a bit more defensive. LMK if that works. And yes, I've been getting some errors from GCC lately while using this in a project (such as `include/utarray.h:124:9: error: comparison of unsigned expression < 0 is always false [-Werror=type-limits]`).

troydhanson:
I'm just thinking it over. All those casts make the macros harder to read. I suppose casting in the place the macro is called _from_ would be another route. Maybe that is just passing the buck. :)

Quuxplusone:
As this PR is four years old, I'm going to be bold and close it (without merging the proposed patch). Please reopen it if this is still an issue for you. And if so, please update the patch to **include a new test** which demonstrates the problem. I just now tried compiling the contents of `tests/` with `-Wtype-limits` and I didn't see any warnings; so either the compiler's behavior has changed in the past 4 years, or I'm using the wrong compiler, or (most likely of all) we have no tests for the behavior you were seeing and so some tests should be added.

PR #75, opened 2016-06-27
Hashvalue v2

Quuxplusone:
@troydhanson, please hold off on merging this until we can get all the required details hashed out.
For example:

- there should be a unit test
- documentation needs regenerating (I can't do that myself)
- come to agreement with @thiloschulz about the best name for the `_BY_HASH_VALUE` methods: is `_BY_HVAL` preferable, or (IMHO) less readable with no obvious upside?

### DIFF
--- a/src/uthash.h
+++ b/src/uthash.h
@@ -105,8 +105,7 @@ typedef unsigned char uint8_t;
 
 #define HASH_VALUE(keyptr,keylen,hashv)                                          \
 do {                                                                             \
- unsigned _hf_bkt;                                                               \
- HASH_FCN(keyptr,keylen,1,hashv,_hf_bkt);                                        \
+ HASH_FCN(keyptr, keylen, hashv);                                                \
 } while (0)
 
 #define HASH_FIND(hh,head,keyptr,keylen,out)                                     \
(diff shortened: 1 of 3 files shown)


thiloschulz:
> hashed out

:-)

thiloschulz:
On second thought, why keep auto-generated files in the git tree at all?

Quuxplusone:
> Just install asciidoc.

I have done that; but running it generates these *huge* diffs compared to what's in the repo, which I can't be bothered to go through and fix up when there's the simple alternative of just asking Troy to run it instead. :)  From your #71 it looks like you have the same problem (e.g. your version of asciidoc changed all the `<tt>` tags to `<code>`).

thiloschulz:
Yes. I didn't bother "fixing" anything because that's what you get when you add auto-generated files into the git repository. I rather suggest the .html files be deleted.

troydhanson:
Ok as you saw on #71 this is incorporated into 650214c. I think we can close this one and you can open a new pull request when you have a unit test for the HVAL stuff. I wrote a minimal note about HASH_VALUE and HVAL in the userguide.

PR #80, opened 2016-06-30
Keeping it simple on the newly added macros (SADD, BY_HVAL)

Quuxplusone:
I'd still like to rename these relatively new API macros with better names; i.e. to get a complete set of

    HASH_FIND[_INT|_STR][_BYHASHVALUE]
    HASH_ADD[_KEYPTR|_INT|_STR][_BYHASHVALUE][_INORDER]
    HASH_REPLACE[_KEYPTR|_INT|_STR][_BYHASHVALUE][_INORDER]

but this is not that PR. This is just a PR to clean up the existing implementations, with no API changes at all (except to stop exposing `HASH_GHV_*`, `HASH_ADD_*LOGUE`, and `HASH_SORTED_ADD` to user code). The benefit is that with the code simplified, it's easier to see how each function is meant to behave and to see which ones are "convenience wrappers" for more primitive functions.

### DIFF
--- a/src/uthash.h
+++ b/src/uthash.h
@@ -224,7 +224,7 @@ do {                                                                            
   (head)->hh.tbl->tail = &((add)->hh);                                           \
 } while (0)
 
-#define HASH_ADD_PROLOGUE(hh,head,keyptr,keylen_in,add)                          \
+#define HASH_SORTED_ADD(hh,head,keyptr,keylen_in,hashval,add,cmpfcn)             \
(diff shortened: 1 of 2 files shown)


thiloschulz:
I think your proposed naming concept is sound and probably better than what we have right now.

Except of course I would use BYHVAL as these names tend to get very long already.

Quuxplusone:
Merged bc910acb678a9f3aeeec1b1d9e0d7d0964a8c349.

PR #81, opened 2016-06-30
Remove generated .html files from the repository.

Quuxplusone:
Running "make all" in the docs/ directory will bring them back.
Also, they're tracked in the gh-pages branch, since that's where
the live documentation on http://troydhanson.github.io/uthash/
comes from.


troydhanson:
Sounds good, I meant to do this, but apparently I forgot to remove the html files other than userguide.

Quuxplusone:
Merged 1d7cb68395a373b4dd2a54a0c1fd13d0fec4fe0d.

PR #82, opened 2016-07-04
In HASH_FIND_IN_BKT, check for hash equality before key equality.

Quuxplusone:
Key equality (HASH_KEYCMP) is generally an expensive memcmp, whereas
if the two keys have different hashes, we can find that out in a single
integer comparison.

This is a particularly big win if your hash table stores long strings
all with the same length; for example, strings that started their life
as SHA hashes.


Quuxplusone:
Merged 04be20500b6a7323effda3fd996315f4c7d6e4ec.

PR #85, opened 2016-07-12
Support targets without standard functions

veselov:
Hi Arthur,

I've been using uthash for a while now, and recently had to port some of my code to an RTOS (Micrium to be specific). Apparently, these things don't quite often provide even standard headers (like stdio), and have a completely alien set of both headers and functions available.

uthash has some portability, as one can override what is used for malloc and etc., but it does insist on including certain headers, and some functions like strlen/memcmp.

I've provided this PR so that one can use uthash on such platforms.

Thank you,
  Pawel.

### DIFF
--- a/tests/keystat.c
+++ b/tests/keystat.c
@@ -107,11 +107,11 @@ int main(int argc, char *argv[])
         percent = atoi(argv[2]);
         v = 3;
     }
-    if ((argc >= v) && (strcmp(argv[v],"-v") == 0)) {
+    if ((v < argc) && (strcmp(argv[v],"-v") == 0)) {
         verbose=1;
         v++;
     }
-    if (argc >= v) {
+    if (v < argc) {
         filename=argv[v];
     }
     fd=open(filename,MODE);
@@ -253,4 +253,3 @@ readmore2:
     }
     return 0;
 }
-


doc/userguide.txt
```diff
@@ -1536,6 +1536,38 @@ You don't need to use these hooks- they are only here if you want to modify
 the behavior of uthash.  Hooks can be used to change how uthash allocates 
 memory, and to run code in response to certain internal events. 
 
+Preventing stadnard includes
```
Quuxplusone: standard

doc/userguide.txt
```diff
@@ -1536,6 +1536,38 @@ You don't need to use these hooks- they are only here if you want to modify
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+For some targets, standard headers may not be available. To prevent uthash to
+pick up such headers, define `uthash_noheaders` before including uthash.h
+
+    #define uthash_noheaders
+    #include "uthash.h"
+
+Memory search and compare
```
Quuxplusone: Actually, I think all this documentation should be folded into the following section on avoiding malloc/free. I can refactor the docs when I pull, though, so you don't actually have to worry about it.

src/uthash.h
```diff
@@ -26,9 +26,11 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #define UTHASH_VERSION 2.0.1
 
+#ifndef uthash_noheaders
```
Quuxplusone: This macro should be ALL_CAPS; I think I'd go with `UTHASH_NO_STANDARD_HEADERS` or else `UTHASH_FREESTANDING_IMPLEMENTATION` depending on your intended semantics.

You still `#include <stdint.h>` below, even when this macro is defined. Is that okay? Regardless, is <stdint.h> present on your platform? Regardless, is <stdint.h> guaranteed to be present on every platform, even freestanding/non-hosted implementations?

src/uthash.h
```diff
@@ -87,6 +89,17 @@ typedef unsigned char uint8_t;
 #ifndef uthash_free
 #define uthash_free(ptr,sz) free(ptr)     /* free fcn                        */
 #endif
+#ifndef uthash_strlen
+#define uthash_strlen(s) strlen(s)        /* STRLEN(3)                       */
+#endif
+#ifndef uthash_memcmp
+#define uthash_memcmp(m1,m2,s) memcmp(m1,m2,s) /* MEMCMP(3)                  */
```
Quuxplusone: Nit: Please remove the comments on lines 93, 96, and 99.

src/uthash.h
```diff
@@ -342,17 +356,17 @@ do {
...
         if ((delptr)->hh.prev != NULL) {                                         \
-            ((UT_hash_handle*)((ptrdiff_t)((delptr)->hh.prev) +                  \
+            ((UT_hash_handle*)((uthash_ptrdiff_t)((delptr)->hh.prev) +           \
                     (head)->hh.tbl->hho))->next = (delptr)->hh.next;             \
         } else {                                                                 \
             DECLTYPE_ASSIGN(head,(delptr)->hh.next);                             \
         }                                                                        \
         if (_hd_hh_del->next != NULL) {                                          \
-            ((UT_hash_handle*)((ptrdiff_t)_hd_hh_del->next +                     \
+            ((UT_hash_handle*)((uthash_ptrdiff_t)_hd_hh_del->next +              \
```
Quuxplusone: My kneejerk reaction here is "why on earth does this cast exist in the first place"; I'd like to dig a little deeper here and figure out whether we can get away with a simpler type, such as `unsigned` or `void*` or `char*` or no-cast-at-all.

I'm inclined to accept the uthash_strlen, uthash_memcmp stuff immediately and defer the ptrdiff_t stuff to another PR; what do you think?

tests/test6.c
```diff
@@ -1,15 +1,27 @@
...
 #define uthash_free(ptr,sz) alt_free(ptr)
+#define uthash_strlen(s) alt_strlen(s)
+#define uthash_memcmp(v1,v2,s) alt_memcmp(v1,v2,s)
+
+static void * search_key = 0;
+static int seen_strlen = 0;
 
 typedef struct example_user_t {
-    int id;
+    char key[3];
```
Quuxplusone: What would go wrong if you didn't change this struct definition (or most of the code below)?  I seem to recall that uthash always uses `memcmp`, even to compare ints.

Quuxplusone:
I'm favorably disposed to this patch in general, but I think it can be simplified, and I'd like to defer the `ptrdiff_t` part to a second patch where the patcher (whether it's yourself or myself or someone else) has thoroughly investigated the role of `ptrdiff_t` in uthash. Please fix some things, squash-rebase and I'll take another look!

tests/test6.c
veselov: Without this change, there is no call to `strlen`

src/uthash.h
veselov: I'll look into this.

src/uthash.h
veselov: Including stdint.h is not OK :) I don't know how I missed it, I'll take another look. As to the naming, I followed the names used for overriding function calls, like `uthash_xxx`. Is the rule here - functions - lower case names, others - upper case? I assumed hooks are lower case.

veselov:
I'll work on this tonight.

src/uthash.h
veselov: I only added them because the other defines had them, but sure.

src/uthash.h
Quuxplusone: Macros in general should be ALL_CAPS; uthash_malloc and uthash_free are lowercase because they're supposed to act like (and therefore look like) malloc and free. So uthash_memcmp and uthash_strlen should also look like memcmp and strlen respectively; but UTHASH_FREESTANDING_IMPLEMENTATION should be uppercase by default because there's no reason to make it not-uppercase.

Re the name, I *think* I like `UTHASH_FREESTANDING_IMPLEMENTATION`, but maybe that conflicts with your statement that `<stdint.h>` isn't available on your platform. http://stackoverflow.com/questions/13710101/the-iso-c-standard-defines-two-classes-of-conforming-implementation claims that `<stdint.h>` is available on freestanding implementations as of C99 (but possibly not before).

tests/test6.c
Quuxplusone: Aha. Okay, then.

Tangent: I assume that none of the test cases compile on your target platform because they *all* use stdio, is that right? Do you have any interest in trying to get the test cases to build on your platform, and/or ideas about what that would look like?

veselov: The platform I am specifically working with is Micrium on top of some Motorola CPU. There will substantial amount of work to get the tests in there, as they have to be converted into a single executable that is then rolled into a flash image. Since we haven't figured out how to even get debug output from our dev board, that will need to be converted to network calls. So, it's possible, but is a lot of work :)

src/uthash.h
veselov: I'll rename it to free-standing, because that's what it's supposed to be. However, I will still snip out `<stdint.h>` (I'll have to check what out of stdint does uthash need though), as this can be documented and developer can include it if they need to. As my understanding goes so far, the world of embedded things don't follow the standards all too well. For example, Micirum has functions like Str_Len() and such, I'm not sure is it because they don't trust the compiler builtins or what not. At the same time, their headers override standard types left and right, so including things like `<stddef.h>` just generates a slough of warnings, and at worse may result into compiler referencing something from builtin library that has ends in libc, which in turn fails linking.

src/uthash.h
```diff
@@ -573,18 +602,22 @@ do {
...
 #if (defined(__GNUC__) && defined(__i386__)) || defined(__WATCOMC__)             \
-  || defined(_MSC_VER) || defined (__BORLANDC__) || defined (__TURBOC__)
-#define get16bits(d) (*((const uint16_t *) (d)))
+  || defined(_MSC_VER) || defined (__BORLANDC__) || defined (__TURBOC__)         \
+  || defined(uthash_uint16_t)
+#ifndef uthash_uint16_t
+#define uthash_uint16_t uint16_t
+#endif
+#define get16bits(d) (*((const uthash_uint16_t *) (d)))
 #endif
```
veselov: I thought for a while on what needs to happen here. Apparently, not using standard includes doesn't define __I386__ so on Linux, this path is not taken. I figure if the caller defined uthash_uint16_t, we should always use that, hence the update.

src/uthash.h
```diff
@@ -593,17 +626,17 @@ do {
   /* Main loop */                                                                \
   for (;_sfh_len > 0U; _sfh_len--) {                                             \
     hashv    += get16bits (_sfh_key);                                            \
-    _sfh_tmp  = ((uint32_t)(get16bits (_sfh_key+2)) << 11) ^ hashv;              \
+    _sfh_tmp  = ((uthash_uint32_t)(get16bits (_sfh_key+2)) << 11) ^ hashv;       \
     hashv     = (hashv << 16) ^ _sfh_tmp;                                        \
-    _sfh_key += 2U*sizeof (uint16_t);                                            \
+    _sfh_key += 2U* 2/*sizeof(uint16_t)*/;                                       \
```
veselov: There is this and another instance where I replaced sizeof(uint16_t) with 2. Now, AFAIU, sizeof would not have worked if uint16_t is not defined, so this should've failed on exotic platforms anyway. And sizeof(uint16_t) must always be 2. If the chosen get16bits() implementation is using byte assembling, then there is still no uint16_t (or uthash_uint16_t).

veselov:
I modified the PR. Let me know if there are still other concerns. I don't know how to squash this on the client side, I think I made more mess that needed, but the combine commits do what they should.

Quuxplusone:
FYI, to squash on your side, you can just do

    git fetch
    git checkout master
    git rebase -i 539b450   # or "git rebase -i upstream/master", if you have this repo set as a remote named "upstream"
    git push -f origin master

The reason this is ugly is because you're not working in a feature branch. If you were working in a feature branch, it would look like

    git fetch
    git checkout your-branch
    git rebase -i origin/master   # because origin/master won't have moved
    git push -f origin your-branch

FYI, I'm working right now on adopting your basic idea of uthash_memcmp and uthash_strlen. All this `uthash_intXX_t` stuff got real complicated real fast and I think I'm going to hold off on that part in the interests of Keeping It Simple.

(Looks like some of this code is overcomplicated due to a few PRs such as #50 whose intent was to make the code MISRA-clean. Personally, I doubt that was a worthwhile or long-term-maintainable goal.)

troydhanson:
I'm sure you're right about the linting business, it made everything harder to read.

src/uthash.h
```diff
@@ -26,9 +26,11 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #define UTHASH_VERSION 2.0.1
 
+#ifndef UTHASH_FREESTANDING_IMPLEMENTATION
 #include <string.h>   /* memcmp,strlen */
 #include <stddef.h>   /* ptrdiff_t */
 #include <stdlib.h>   /* exit() */
```
Quuxplusone: My next rabbit-hole question here is: if you skip including these headers, then where do you get the `NULL` macro from? Is there a more freestanding-friendly way for us to get it?

FYI I'm still looking into the `uint8_t`, `uint32_t`, `ptrdiff_t` question, but my gut instinct is that I *don't* want to adopt your proposed `uthash_uint8_t` stuff; it seems really un-simple.

veselov: You get NULL from somewhere.

As an alternative to providing configuration macros for all of `XXX_t`s, uthash may simply demand that they are present; and list all things it demands are present in free standing configuration.

That statement about you not taking a patch makes sense. I consider `NULL` a standard (something that better be defined somehow), but `XXX_t` not so standard, and instead of requiring the caller to provide them, giving the caller a way to hook. As long as it's done in one way or another, I'm actually fine (for embedding), I just don't know what the litmus test for those things is. A developer may want to override `malloc()` because they want a different memory allocation, but it also may be because they have no `malloc()`. May be the litmus test should be: _"Is it conceivable that THING exists, but does/means something else that the developer might want it to"_. Like, overriding a `malloc()`, and for uthash only, is possible, but just a pain. Having uuin32_t that doesn't mean "unsigned 32-bit integer" is not conceivable.

PR #88, opened 2016-07-23
some lintings

fperrad:


src/uthash.h
```diff
@@ -119,7 +119,7 @@ do {
 #define HASH_FIND_BYHASHVALUE(hh,head,keyptr,keylen,hashval,out)                 \
 do {                                                                             \
   (out) = NULL;                                                                  \
-  if (head) {                                                                    \
+  if (head != NULL) {                                                            \
```
Quuxplusone: This change is wrong, because `head` is a macro parameter that needs parentheses whenever it's an operand.
(I would welcome a patch that fixed any lingering issues in that area; I'm sure there are some. However, that patch would suffer from the same bit-rot problem as all these "linting" patches: the code would quickly fall out of conformance unless the patch actually added test cases to the test suite exercising the new invariants.)

src/uthash.h
```diff
@@ -245,12 +245,14 @@ do {
...
         break;                                                                   \
-    } while ((_hs_iter = _hs_iter->next));                                       \
-    if (_hs_iter) {                                                              \
+      }                                                                          \
+    } while ((_hs_iter = _hs_iter->next) != NULL);                               \
+    if (_hs_iter != NULL) {                                                      \
       (add)->hh.next = _hs_iter;                                                 \
-      if (((add)->hh.prev = _hs_iter->prev)) {                                   \
+      (add)->hh.prev = _hs_iter->prev;                                           \
+      if ((add)->hh.prev != NULL) {                                              \
```
Quuxplusone: I'm amenable to this change, although it doesn't seem particularly useful to me. I'd prefer a patch with a commit message such as "uthash produces the following diagnostics on platform Foo: (bar)... Suppress these diagnostics by consistently (baz)..."

And if that patch could consist entirely of `#ifdef Foo ... #pragma warning suppress bar ... #endif` somewhere at the top of this header, then that would be even better, because it wouldn't penalize anybody who didn't use platform Foo.

src/uthash.h
```diff
@@ -719,7 +721,7 @@ do {
     (out) = NULL;                                                                \
   }                                                                              \
   while ((out) != NULL) {                                                        \
-    if ((out)->hh.hashv == (hashval) && (out)->hh.keylen == (keylen_in)) {       \
+    if (((out)->hh.hashv == (hashval)) && ((out)->hh.keylen == (keylen_in))) {   \
```
Quuxplusone: This change seems harmless but at the same time I'm not in favor of it because it seems to complicate this expression for no actual gain.

If this is to deal with some MISRA-C rule, could you provide a pointer to some code that uses uthash and needs to compile with MISRA-C rules? Again I feel like the proper workaround might just be to suppress that diagnostic in files that `#include "uthash.h"`.

PR #90, opened 2016-07-29
Make the utlist.h unit tests pass cleanly with `NO_DECLTYPE`.

### LINKED ISSUE #89: utlist - Missing VS2008 versions of several defines
Compiling as C in Visual Studio breaks.  `NO_DECLTYPE` is defined, but several "functions" are not redefined to use VS2008 versions.  Many of those "functions" still attempt to use `LDECLTYPE`, which is defined as `char*`.  This leads to compile errors for assigning a `struct*` to a `char*` or trying to access `(char*)->next`.

Short example:
```
#include <stdlib.h>
#include "utlist.h"

typedef struct mystruct {
	struct mystruct *next;
	int someValue;
} mystruct;

void testutlist()
{
	mystruct *head = NULL;
	mystruct *elem1 = (mystruct*)malloc(sizeof(mystruct));
	mystruct *elem2 = (mystruct*)malloc(sizeof(mystruct));

	LL_APPEND(head, elem1);
	LL_PREPEND_ELEM(head, elem1, elem2);
}
```

Two compile errors on the LL_PREPEND_ELEM line:
```
1>c:\...\source.c(16): warning C4133: '=' : incompatible types - from 'mystruct *' to 'char *'
1>c:\...\source.c(16): error C2223: left of '->next' must point to struct/union
```
are actually due to utlist.h lines 493 & 494, because `_tmp` was declared as a `char*`.

Quuxplusone:
After this patch, the following command works fine:

    cd tests ; CFLAGS='-D_MSC_VER=1' make pedantic

There are probably some functions in utlist.h that still don't work
in VS2008 mode, but if so, we don't unit-test them.

Fixes #89.

TODO: Add a test case based on tinydtls' use of this header.

### DIFF
--- a/src/utlist.h
+++ b/src/utlist.h
@@ -68,11 +68,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define LDECLTYPE(x) decltype(x)
 #else                     /* VS2008 or older (or VS2010 in C mode) */
 #define NO_DECLTYPE
-#define LDECLTYPE(x) char*
 #endif
 #elif defined(__ICCARM__)
 #define NO_DECLTYPE
-#define LDECLTYPE(x) char*
 #else                      /* GNU, Sun and other compilers */
 #define LDECLTYPE(x) __typeof(x)
 #endif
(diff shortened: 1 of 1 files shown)


NickSchaf:
Works in my test app and compiles into tinydtls using VS2008+.

NickSchaf:
Rather than keeping versions of the defines with and without `_tmp` declarations, wouldn't it be cleaner to use only the versions without `_tmp` since they work whether or not `decltype` is available?
Example: replace the original definition of LL_APPEND2 (line 339-350) with the redefined version (494-504) and avoid the redefinition altogether.

Quuxplusone:
I predict that the versions without `_tmp` (e.g. the version of LL_APPEND2 that uses `add->next` as scratch space) are less efficient than the ones that are able to use `_tmp`. In the case of LL_APPEND2, if we have `_tmp` then we don't have to keep dereferencing `add` every time through the tight inner loop.

In the case of LL_CONCAT2, it doesn't much matter whether we use `_tmp` or `(head1)` — but it's still true that the latter *might* be a more expensive expression (e.g. `(*ptr)`), and the latter *definitely* has a larger lifetime than the lifetime of `_tmp`, which could affect codegen. Therefore, in the cases where we have the *ability* to use a narrowly scoped temporary variable `_tmp`, IMO we *should* use it.

Quuxplusone:
Merged 8052cb8ce46230a5c3d9d3ea84798b69515899af, and the tinydtls-derived test case in 24a6c0c1bf545729d218d52b865319b3970dd8ab.

PR #95, opened 2016-11-25
Add utstring_release

tbeu:
Not sure if you want to have it as function or macro.

### DIFF
--- a/src/utstring.h
+++ b/src/utstring.h
@@ -121,6 +121,15 @@ do {                                                             \
 
 #define utstring_body(s) ((s)->d)
 
+static char* utstring_release(UT_string *s) {
+  char* ret = NULL;
+  if (s != NULL) {
+    ret = s->d;
+    free(s);
+  }
+  return ret;
+}
+
 _UNUSED_ static void utstring_printf_va(UT_string *s, const char *fmt, va_list ap) {
    int n;
    va_list cp;


src/utstring.h
```diff
@@ -121,6 +121,15 @@ do {                                                             \
 
 #define utstring_body(s) ((s)->d)
 
+static char* utstring_release(UT_string *s) {
```
Quuxplusone: If this isn't going to be a macro, I think it needs the word `_UNUSED_` in front of it.
No comment on the functionality; I'll let @troydhanson take it or leave it.

troydhanson: I have done this kind of thing before manually. So, I think it's a useful operation. Agreed on the need for `_UNUSED_`. I think the name `_release` is so-so. I want to imply that it's releasing it "to" the caller. Like `utstring_disown`. Thoughts? Incidentally I once or twice wrote the opposite function that adopts the string you give it and makes a utstring around it. From that point of view, `utstring_own` and `utstring_disown` are a pair.

Quuxplusone: Surely the opposite of `utstring_disown` would be `utstring_adopt`!

The longer I look at the existing utstring interface, the more I think it'll never quite line up with my C++ish expectations. Seems like I get a utstring by calling `utstring_new(p)`, but then to put a string into it I have to use `utstring_bincpy(p, "mydata", strlen("mydata"))` (there's no `utstring_strcpy`), and then when I'm done I do `utstring_free(p); p = NULL;` (if I forget that reset-to-null operation, I'll get a double-free if I `utstring_renew(p)` — I think `utstring_renew` is an unusual wart in the API).

With this patch, if I want to "release/disown" the string, I do either `char *s = utstring_body(p); utstring_init(p);` or else `char *s = utstring_release(p); p = NULL;` (again forgetting the reset-to-null might lead to double-frees) — and if I want to "release/disown" a non-alloc'ed utstring, I have to use the first method, because `utstring_release` always wants to `free` the utstring I give it.

I'd want to separate the act of "disowning the buffer" and the act of "destroying the (possibly non-alloc'ed) utstring object itself":

```
_UNUSED_ static char *utstring_disown(UT_string *s) {
  char *ret = s->d;
  s->n = 0; s->i = 0; s->d = NULL;
  return ret;
}
```

tbeu:
So what about `utstring_atach` and `utstring_detach` as names?

troydhanson:
I think disown/adopt is more intuitive (well, that's an overstatement, somewhat more intuitive). About the renew/new/free - it never occurred to me to have something freed and then renewed. The way I've used renew is to delay initialization of a string until it's needed, then I leave it for the lifetime of a program and then only free it at the end. If you think free-and-later-renew is a use case that needs to work better we could add `s=NULL` to the utstring_free macro.

I'm still pondering the idea of separating the acts above, my intuition was that the original way (release does both) made sense. I need to turn my attention to other things this week so I trust your judgment to make a decision.

Quuxplusone:
@troydhanson: My intuition is "I don't really care about utstring, and don't have a good mental model for what the primitive operations are *supposed* to be (see my comment above); so, I'm just going to let this patch languish until @troydhanson has time to figure it out." :)

I agree that disown/adopt is the best pair of verbs. IMO @tbeu should adopt those verbs and provide macros (or `_UNUSED_ static` functions) for both verbs in the same patch.

> If you think free-and-later-renew is a use case that needs to work better

Well, that's the thing: I currently have no mental model for what use-cases are *supposed* to work and which aren't. (I'll pick this back up in the comment thread above.)

troydhanson:
I agree with the recommendation for those verbs and providing both in one patch. Let us know if you're up for that @tbeu. :)

tbeu:
Sure, I'll revise this PR accordingly.

tbeu:
a4f1104 adds `utstring_adopt`/`utstring_disown` as macros.

doc/utstring.txt
```diff
@@ -154,16 +154,18 @@ Operations
 | utstring_new(s) | allocate a new utstring
 | utstring_renew(s) | allocate a new utstring (if s is `NULL`) otherwise clears it
 | utstring_free(s) | free an allocated utstring
-| utstring_init(s)  |  init a utstring (non-alloc)
-| utstring_done(s) |  dispose of a utstring (non-allocd)
+| utstring_init(s)  | init a utstring (non-alloc)
+| utstring_done(s) | dispose of a utstring (non-alloc)
```
Quuxplusone: Thanks! I'll take these diffs to the existing docs upstream right now (hopefully in a way that merges cleanly with this PR).

src/utstring.h
```diff
@@ -121,6 +121,24 @@ do {                                                             \
...
+    (s)->n = (l);                                          \
+    (s)->i = (l);                                          \
+  }                                                        \
+} while(0)
+
+#define utstring_disown(s,b)                               \
+do {                                                       \
+  if (s) {                                                 \
+    (b) = (s)->d;                                          \
+    utstring_clear(s);                                     \
```
Quuxplusone: This is incorrect; `utstring_clear` will set `s->d[0] = '\0'`, blowing away the contents of `b`. What you want here is more like `utstring_init(s)`, but actually calling out to `utstring_init` seems confusing to me; I think you should just inline the couple of statements you want here. Which IMO would be

    (b) = (s)->d;
    (s)->d = NULL;
    (s)->i = 0;
    (s)->n = 0;

src/utstring.h
```diff
@@ -121,6 +121,24 @@ do {                                                             \
 
 #define utstring_body(s) ((s)->d)
 
+#define utstring_adopt(s,b,l)                              \
+do {                                                       \
+  if (s) {                                                 \
+    (s)->d = (b);                                          \
+    (s)->n = (l);                                          \
+    (s)->i = (l);                                          \
```
Quuxplusone: @troydhanson, in the codebases where you've needed this function in the past, would this have sufficed? Or should it be `utstring_adopt(s,d,i,n)` for cases where the initial capacity is intentionally greater than the initial length?
I guess since C's `realloc` and `free` don't need to know the original capacity, this version is safe; but if someone later filed a feature request to make utstring work with `ptr = sizedmalloc(sz)` and `sizedfree(ptr,sz)` the way uthash currently does, it would be hard to retrofit that into this "length-only" model.

src/utstring.h
tbeu: Yes, sorry, I changed it to `utstring_clear ` after I tested which was wrong.

tbeu:
Ready to merge?

Quuxplusone:
I'd still like to get @troydhanson's feedback on my comment about the possibility of `capacity != length` https://github.com/troydhanson/uthash/pull/95/files#r90346232 — but if there's nothing forthcoming in the next 48 hours, then sure, I'll squash and merge this patch as is.

src/utstring.h
troydhanson: It needs a tweak. The other "put" functions (`utstring_printf`, `utstring_bincpy`) internally set the internal buffer length `n` one larger than the input size, to accommodate a trailing nul, and then forcibly put it there. (Look at the `utstring_bincpy` definition to see what I mean).
It's a convenience so there is a guaranteed nul terminator in the spare capacity part of the string. It is a quirk of utstring that I like because, it makes it compatible with printf (etc) that expect nul-terminated strings, while keeping the nul past the length `i` means that the content/length are unharmed.

One way to preserve that notion, is to switch `adopt` to a two parameter version `utstring_adopt(s,d)` that demands a C string as input, and then its implementation sets 
<pre>
s->i = strlen(d);
s->n = s->i +1 ;
</pre>

The problem with the current three-parameter adopt is that the caller can pass a non-terminated string, breaking the expectation that utstring always has a `nul` after the buffer. 

We can switch to the four-parameter version `utstring_adopt(s,d,i,n)` but, to preserve the `nul` expectation, I would explicitly force a nul at d[i] (which requires a preceding realloc if i==n).

tbeu:
Anything more I can do here?

tbeu:
> but if there's nothing forthcoming in the next 48 hours, then sure, I'll squash and merge this patch as is.

It's almost 48 months now.

src/utstring.h
Quuxplusone: @tbeu: Sorry! I think what happened here was that the feedback I was awaiting on 2016-12-15 *did* arrive, in the form of Troy's comment on 2016-12-16. And then I never explicitly said "@tbeu please address this," but I also never really bothered (AFAIR) to figure out whether I even agreed with what Troy said, either.

I suggest that at this point, with many outdated review comments and some pieces of this patch merged already, the best thing to do would be to close this PR and (if you still want the feature merged) open a fresh PR.

*Do* you have any thoughts on the desirability of `utstring_adopt(s, buf, len)` versus `utstring_adopt(s, ntbs)` versus `utstring_adopt(s, buf, len, cap)`?  I think that (at the moment) I agree with Troy that `utstring_adopt(s, buf, len)` is the worst of both worlds, because it would let you create a non-null-terminated utstring.

PR #97, opened 2016-12-17
unsigned casting in utstring_len macro cause truncation on 64 bit (ch…

mse11:
…anged to size_t). Range of 'unsigned' is half of 'size_t'. sizeof(unsigned)4 vs. sizeof(size_t)8.

### DIFF
--- a/src/utstring.h
+++ b/src/utstring.h
@@ -117,7 +117,7 @@ do {                                                             \
   (dst)->d[(dst)->i]='\0';                                       \
 } while(0)
 
-#define utstring_len(s) ((unsigned)((s)->i))
+#define utstring_len(s) ((size_t)((s)->i))
 
 #define utstring_body(s) ((s)->d)
 
--- a/src/utstring.h
+++ b/src/utstring.h
@@ -117,7 +117,7 @@ do {                                                             \
   (dst)->d[(dst)->i]='\0';                                       \
 } while(0)
 
-#define utstring_len(s) ((size_t)((s)->i))
+#define utstring_len(s) ((s)->i)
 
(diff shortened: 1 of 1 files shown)


src/utstring.h
```diff
@@ -117,7 +117,7 @@ do {                                                             \
   (dst)->d[(dst)->i]='\0';                                       \
 } while(0)
 
-#define utstring_len(s) ((unsigned)((s)->i))
+#define utstring_len(s) ((size_t)((s)->i))
```
mse11: You are right - thanks

troydhanson: I dunno why it was there either. Probably someone quieted a warning (maybe on 32 bit system) that way.

Quuxplusone: Oho, now I know what'll break if we remove it.

(1) Many test cases call `printf("%u", utstring_len(s))`; this needs to change to `printf("%zu", utstring_len(s));`, and then we lose C89 compatibility because `%zu` is a C99 invention. The proper workaround there is probably to add back the cast: `printf("%u", (unsigned)utstring_len(s))`. We need to do *something* to fix the warnings from the test suite, anyway.

(2) In C++, the type of an expression affects overload resolution; e.g. `std::max(100u, utstring_len(s))` will no longer compile. However, IMHO nobody should even *think* about using utstring in C++ given that `std::string` is available! So I'm willing to punt on this one.

I'll go ahead and merge this patch with the above fix to issue (1), unless someone objects first.

Quuxplusone:
Merged 85f3db073c92654d296d9814b5bc5b807c53a8dc — thanks!

PR #98, opened 2016-12-29
Fix Author-string for older versions of asciidoc

besser82:
Older versions of Asciidoc fail with `malformed author`.

### DIFF
--- a/doc/userguide.txt
+++ b/doc/userguide.txt
@@ -1,6 +1,6 @@
 uthash User Guide
 =================
-Troy D. Hanson, Arthur O'Dwyer
+:Author: Troy_D._Hanson_and_Arthur_O'Dwyer
 v2.0.1, July 2016
 
 To download uthash, follow this link back to the


Quuxplusone:
Can you cut-and-paste the whole terminal output, plus the output of `asciidoc --version`, for this old version? I'm on `asciidoc 8.6.9` myself, and this patch results in a generated userguide.html containing literal underscores (i.e. they're not turning into spaces or whatever you expected to happen). Do you think the problem could be related to the `'` in my last name? (I get that a lot. ;))

besser82:
The error occurs with `asciidoc-8.4.5-4.1.el6` on Red Hat Enterprise Linux 6 / CentOS 6.  I've fixed the problem with the underscores not changed to whitespace, in fact asciidoc didn't do it because of the `,` inside the string.  Now everything should work fine.

besser82:
Well, don't mind.  It seems not to work always…  I'll keep this a downstream-patch for el6 as long as it is needed.

PR #100, opened 2017-01-11
Fix #99: segfault in HASH_ADD_KEYPTR_INORDER.

### LINKED ISSUE #99: Possible bug in HASH_ADD_KEYPTR_INORDER causing segfault
Greetings.
I've been using uthash for years but recently it seems I found a bug in it. 
Bug is in pointer keys with sorting function, i.e. in HASH_ADD_KEYPTR_INORDER. And it causes segfault.
It seems strange, but it really looks like a bug. I read documentation some times and I guess I used it properly, but it failed. 

I made a small test case to show the bug. 
I also sketched a patch (only for GCC, since I don't have other compilers at hand, but it's easily extendable for other declarations).
I uploaded the test case and patch here:
https://github.com/IronBug/uthash_test

Sincerely, Yana A. Kireyonok aka Iron Bug

Quuxplusone:
Thanks to @IronBug for the bug report (#99) and test90.c!

test87.c would have caught this bug at the time the macros were added,
except that by chance `offsetof(hstruct_t, hh)` was `0`. I've adjusted
the test so that it fails before this patch and succeeds after.

### DIFF
--- a/src/uthash.h
+++ b/src/uthash.h
@@ -242,20 +242,20 @@ do {                                                                            
     (head) = (add);                                                              \
     HASH_MAKE_TABLE(hh, head);                                                   \
   } else {                                                                       \
-    struct UT_hash_handle *_hs_iter = &(head)->hh;                               \
+    void *_hs_iter = (head);                                                     \
     (add)->hh.tbl = (head)->hh.tbl;                                              \
     do {                                                                         \
-      if (cmpfcn(DECLTYPE(head) ELMT_FROM_HH((head)->hh.tbl, _hs_iter), add) > 0) \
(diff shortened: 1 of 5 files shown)


thiloschulz:
Not true, test87.c already tests the INORDER macros.

Quuxplusone:
Okay, this patch is updated with more thorough tests (I just adjusted test87.c) and ready to merge, which I'll do circa Monday if I don't hear any objections.

Quuxplusone:
Merged 13723b1c0d16e2a80e0f6bf0175aaadcd6a6c8f5. Thanks again @IronBug and please do submit some complicated test cases for this stuff if you get the chance. :)

PR #101, opened 2017-01-17
Added INORDER macros for inserting into sorted lists

jeffreylovitz:
This proposed commit adds insertion macros that, using a given comparator (ostensibly the one used for the SORT macro) inserts an element into a single, double, or circular linked list at the proper position.

As this is my first contribution to this library, I welcome critiques on its contents, especially in the area of safe encapsulation.


src/utlist.h
```diff
@@ -349,6 +349,26 @@ do {
...
+#define LL_INSERT_INORDER(head,add,cmp)                                                        \
+  do {                                                                                         \
+    LDECLTYPE(head) _pos;                                                                      \
+    _CASTASGN(_pos, head);                                                                     \
+    LL_FIND_POS(_pos,add,cmp);                                                                 \
+    LL_APPEND_ELEM(head,_pos,add);                                                             \
+  } while (0)
+
+#define LL_FIND_POS(head,node,cmp)                                                             \
+  LL_FIND_POS2(head,node,cmp,next)                                                             \
```
Quuxplusone: Nit: please remove the trailing backslash on this line (the last line of the macro definition).

src/utlist.h
```diff
@@ -349,6 +349,26 @@ do {
...
+
+#define LL_FIND_POS2(head,node,cmp,next)                                                       \
+  do {                                                                                         \
+    for (; (head)->next != NULL; (head) = (head)->next ) {                                     \
+      if (cmp((head)->next, (node)) > 0) {                                                     \
```
Quuxplusone: Nit: You don't need the parens around `node` here because it's being passed as an argument (so the only bad interaction it could have is if it had embedded commas, and macro arguments can't have embedded commas). Since the parens are unneeded, please remove them.

(Also note that if you wanted to avoid weird interactions with `cmp` you'd have to write `(cmp)`; but we actually *want* the weird interactions there, e.g. if `cmp` is the name of a two-argument macro then we'd like to expand the macro. So that's correct.)

src/utlist.h
```diff
@@ -600,6 +620,31 @@ do {
...
+    DL_APPEND_ELEM(head,_pos,add);                                                             \
+  } while (0)
+
+
+#define DL_FIND_POS(head,node,cmp)                                                             \
+  DL_FIND_POS2(head,node,cmp,prev,next)                                                        \
+
+#define DL_FIND_POS2(head,node,cmp,prev,next)                                                  \
+  do {                                                                                         \
+    if (cmp((head), (node)) > 0) {                                                             \
```
Quuxplusone: What if `head` is `NULL`? (And please add a test case that covers that possibility.)

src/utlist.h
```diff
@@ -600,6 +620,31 @@ do {
...
+      (head) = NULL;                                                                           \
+    } else {                                                                                   \
+      for (; (head)->next != NULL; head = (head)->next ) {                                     \
```
Quuxplusone: Nit: `head` should be `(head)`

src/utlist.h
```diff
@@ -600,6 +620,31 @@ do {
...
+        if ((head)->next == (node)) break;                                                     \
+        if (cmp((head)->next, (node)) > 0) {                                                   \
```
Quuxplusone: Nit: `(node)` should be `node`, per [the above explanation](https://github.com/troydhanson/uthash/pull/101/files#r96501393).

src/utlist.h
```diff
@@ -349,6 +349,26 @@ do {
   }                                                                                            \
 } while (0)
 
+#define LL_INSERT_INORDER(head,add,cmp)                                                        \
+  do {                                                                                         \
+    LDECLTYPE(head) _pos;                                                                      \
+    _CASTASGN(_pos, head);                                                                     \
+    LL_FIND_POS(_pos,add,cmp);                                                                 \
```
Quuxplusone: I think I'd like to see the phase `FIND_POS` replaced throughout with `LOWER_BOUND`, as in [std::lower_bound](http://en.cppreference.com/w/cpp/algorithm/lower_bound). That will help people remember what exactly it does. (E.g. it does *not* return NULL if the element isn't found; it *does* return specifically the first instance of the element if there are multiple consecutive copies of it; etc.)

src/utlist.h
```diff
@@ -349,6 +349,26 @@ do {
   }                                                                                            \
 } while (0)
 
+#define LL_INSERT_INORDER(head,add,cmp)                                                        \
+  do {                                                                                         \
+    LDECLTYPE(head) _pos;                                                                      \
```
Quuxplusone: Weird, GitHub ate one of my line comments. I'll repeat it: I suspect that this new code won't work with `NO_DECLTYPE`. Can you investigate it a little bit (e.g. by writing some test cases, and then trying to compile those test cases with `-DNO_DECLTYPE=1`) and fix any really low-hanging fruit?
I'm not too worried if there remains fruit that's not low-hanging. Worst case, some list macros will be unusable on compilers with `NO_DECLTYPE`; but then people on those compilers can just avoid those macros and they'll be fine.

jeffreylovitz: @Quuxplusone 
I resolved the head == NULL possibility in the INORDER macros and made the other changes that you advocated for.  I was not able to generate variants of the macros that would work with `NO_DECLTYPE` compilers - the only solution that I see would be passing a dummy node to the macros as an argument.

Please let me know if there are further changes I could make to improve this contribution!

src/utlist.h
```diff
@@ -775,6 +836,37 @@ do {
  (head) = (add);                                                                               \
 } while (0)
 
+#define CDL_INSERT_INORDER(head,add,cmp)                                                       \
```
Quuxplusone: Shouldn't there be a `CDL_INSERT_INORDER2` for the benefit of lists with non-default next and prev fields?

src/utlist.h
```diff
@@ -349,6 +349,35 @@ do {
   }                                                                                            \
 } while (0)
 
+#define LL_INSERT_INORDER(head,add,cmp)                                                        \
```
Quuxplusone: Shouldn't there be an `LL_INSERT_INORDER2` for the benefit of lists with non-default next and prev fields?

tests/Makefile
```diff
@@ -13,7 +13,8 @@ PROGS = test1 test2 test3 test4 test5 test6 test7 test8 test9   \
         test58 test59 test60 test61 test62 test63 test64 test65 \
         test66 test67 test68 test69 test70 test71 test72 test73 \
         test74 test75 test76 test77 test78 test79 test80 test81 \
-        test82 test83 test84 test85 test86 test87 test88 test89
+        test82 test83 test84 test85 test86 test87 test88 test89 \
+				test90 test91 test92
```
Quuxplusone: This merge-conflicts with 13723b1c0d16e2a80e0f6bf0175aaadcd6a6c8f5; please `git rebase -i upstream/master` and fix the conflict. (Also, looks like some tabs-for-spaces issues crept in here; please fix that too.)  To get an `upstream` remote into your local copy of the repo, you can add these lines to your `uthash/.git/config`:

    [remote "upstream"]
        url = git@github.com:troydhanson/uthash
        fetch = +refs/heads/*:refs/remotes/upstream/*

As far as I'm concerned, these three tests could/should be done in a single `test91.c` file, one after the other.

jeffreylovitz:
@Quuxplusone 
I made the changes you suggest.  Let me know if there are further concerns with this branch that I can address!

jeffreylovitz:
@Quuxplusone 
Are there any changes you would like to see made in this code?  I'm happy to improve it to better match the repo if necessary.

src/utlist.h
```diff
@@ -349,6 +349,38 @@ do {
...
+      (head)->next = NULL;                                                                     \
+    }                                                                                          \
+  } while (0)
+
+#define LL_LOWER_BOUND(head,node,cmp)                                                          \
+  LL_LOWER_BOUND2(head,node,cmp,next)
+
+#define LL_LOWER_BOUND2(head,node,cmp,next)                                                    \
+  do {                                                                                         \
+    if (cmp(head, node) > 0) {                                                                 \
```
Quuxplusone: Here and throughout: `> 0` should be `>= 0` so that the following test case passes:

```
struct S { int v; struct S *next; };

int scmp(const struct S *a, const struct S *b)
{
    return (a->v < b->v) ? -1 : (a->v > b->v);
}

int main()
{
    struct S a = { 1, NULL };
    struct S b = { 1, NULL };
    struct S c = { 1, NULL };

    struct S *head = NULL;
    LL_INSERT_INORDER(head, &a, scmp);
    LL_INSERT_INORDER(head, &b, scmp);
    LL_INSERT_INORDER(head, &c, scmp);
    assert(head == &c && c.next == &b && b.next == &a && a.next == NULL);
}
```

Inserting 100 copies of "x" into the same empty list should take O(100) time, not O(100*100) time.
Please add a test case for this, too.

src/utlist.h
```diff
@@ -600,6 +632,41 @@ do {
...
+
+#define DL_INSERT_INORDER2(head,add,cmp,prev,next)                                                        \
+  do {                                                                                         \
+    LDECLTYPE(head) _pos;                                                                      \
+    _CASTASGN(_pos, head);                                                                     \
+    if (head) {                                                                                \
+      DL_LOWER_BOUND(_pos,add,cmp);                                                            \
+      DL_APPEND_ELEM(head,_pos,add);                                                           \
+    }                                                                                          \
+    else {                                                                                     \
```
Quuxplusone: Whitespace nit: please cuddle this else as `} else {`. (Also please fix the misaligned `\` on line 638.)

src/utlist.h
```diff
@@ -600,6 +632,41 @@ do {
...
+      (head) = add;                                                                            \
```
Quuxplusone: `add` should be `(add)`.

src/utlist.h
```diff
@@ -600,6 +632,41 @@ do {
...
+#define DL_LOWER_BOUND(head,node,cmp)                                                          \
+  DL_LOWER_BOUND2(head,node,cmp,prev,next)                                                     \
+
+#define DL_LOWER_BOUND2(head,node,cmp,prev,next)                                               \
+  do {                                                                                         \
+    if (cmp(head, node) > 0) {                                                                 \
+      (head) = NULL;                                                                           \
+    } else {                                                                                   \
+      for (; (head)->next != NULL; (head) = (head)->next ) {                                   \
+        if ((head)->next == (node)) break;                                                     \
```
Quuxplusone: Why is line 663 needed/wanted? The `LL_` and `CDL_` versions of this code don't have this line.

src/utlist.h
```diff
@@ -775,6 +842,40 @@ do {
...
+
+#define CDL_INSERT_INORDER2(head,add,cmp,prev,next)                                                       \
+  do {                                                                                         \
+    LDECLTYPE(head) _pos;                                                                      \
+    _CASTASGN(_pos, head);                                                                     \
+    if (head) {                                                                                \
+      CDL_LOWER_BOUND(_pos,add,cmp);                                                           \
+      CDL_APPEND_ELEM(head,_pos,add);                                                          \
+    } else {                                                                                   \
+      (head) = add;                                                                            \
```
Quuxplusone: `add` should be `(add)`.

src/utlist.h
```diff
@@ -775,6 +842,40 @@ do {
...
+    if (cmp(head, node) > 0) {                                                             \
+      (head) = NULL;                                                                           \
+    } else {                                                                                   \
+      for (; (head)->next != _tmp; head = (head)->next ) {                                     \
+        if (cmp((head)->next, node) > 0) {                                                   \
+          break;                                                                               \
+        }                                                                                      \
+      }                                                                                        \
+    }                                                                                          \
+  } while(0)
```
Quuxplusone: Nit: `} while(0)` should be `} while (0)`. (And as noted above, it should start in column 1, not 3; and please fix the misaligned `\`s throughout.)

src/utlist.h
```diff
@@ -349,6 +349,38 @@ do {
...
+
+#define LL_LOWER_BOUND(head,node,cmp)                                                          \
+  LL_LOWER_BOUND2(head,node,cmp,next)
+
+#define LL_LOWER_BOUND2(head,node,cmp,next)                                                    \
+  do {                                                                                         \
+    if (cmp(head, node) > 0) {                                                                 \
+      (head) = NULL;                                                                           \
+    } else {                                                                                   \
+      for (; (head)->next != NULL; (head) = (head)->next ) {                                   \
```
Quuxplusone: Nit: `next ) {` should be `next) {` here and on lines 662 and 872.

jeffreylovitz:
@Quuxplusone 

I'm sorry it took me so long to get around to this! I think all of your suggested edits were perfectly reasonable; they should all be fixed now. I also changed some variable names to the standards established in the documentation (primarily converting to `elt`).

I added details of the functions to the documentation file; as well as including the code explanation you wrote there. If that's not the appropriate place, please let me know! 

Thanks again!

Quuxplusone:
Hi @jeffreylovitz,

I made a few whitespace and correctness tweaks in my fork. Please take a look at https://github.com/Quuxplusone/uthash/commit/3ec23b4dc7e28e0afc3f5ae2c070b707bf285ea2 and let me know if you see any problems with it. If it looks good to you, I'll merge it and close this PR. :)

src/utlist.h
```diff
@@ -600,6 +631,39 @@ do {
...
+#define DL_INSERT_INORDER2(head,add,cmp,prev,next)                                             \
+do {                                                                                           \
+  LDECLTYPE(head) _tmp;                                                                        \
+  if (head) {                                                                                  \
+    DL_LOWER_BOUND(head,_tmp,add,cmp);                                                              \
+    DL_APPEND_ELEM(head,_tmp,add);                                                             \
+  } else {                                                                                     \
+    (head) = (add);                                                                            \
+    (head)->next = NULL;                                                                       \
+    (head)->prev = NULL;                                                                       \
```
Quuxplusone: FYI: this should be `(head)->prev = (head)` to preserve utlist's invariant that `(head)->prev` in a doubly-linked list points to the tail of the list.

src/utlist.h
```diff
@@ -349,6 +349,37 @@ do {
...
+    (head)->next = NULL;                                                                       \
+  }                                                                                            \
+} while (0)
+
+#define LL_LOWER_BOUND(head,elt,like,cmp)                                                      \
+  LL_LOWER_BOUND2(head,elt,like,cmp,next)
+
+#define LL_LOWER_BOUND2(head,elt,like,cmp,next)                                                \
+  do {                                                                                         \
+    if (cmp(head, like) >= 0) {                                                                \
```
Quuxplusone: FYI: this code doesn't account for the possibility that `(head) == NULL`.

src/utlist.h
```diff
@@ -349,6 +349,37 @@ do {
   }                                                                                            \
 } while (0)
 
+#define LL_INSERT_INORDER(head,add,cmp)                                                        \
+    LL_INSERT_INORDER2(head,add,cmp,next)
+
+#define LL_INSERT_INORDER2(head,add,cmp,next)                                                  \
+do {                                                                                           \
+  LDECLTYPE(head) _tmp;                                                                        \
```
Quuxplusone: FYI: in my version of this patch I make the `NO_DECLTYPE` codepath work as well.
(You can test it with `make pedantic EXTRA_CFLAGS=-DNO_DECLTYPE`.)

tests/test91.c
```diff
@@ -0,0 +1,51 @@
...
+
+typedef struct el {
+  int id, score;
+  struct el *next, *prev;
+} el;
+
+
+static int order_desc(el *a, el *b) {
+  if (a->score < b->score) return 1;
+  return -1;
```
Quuxplusone: FYI: I think it's important to test the "elements with cmp==0" case, so I adjusted this test in my branch.

jeffreylovitz:
Your improvements all look good to me!  Thank you for being so helpful throughout this process!

Quuxplusone:
Merged 47145f5169aa32ad7fde42081138f5b3e13aca62!

PR #108, opened 2017-03-17
Added macros HASH_TOP and HASH_NEXT for dataset-style iteration.

silvioprog:
Hello masters,

This feature would be very, very useful for the ones who want to use uthash as library in other languages that offer `for in` support, like Pascal, JS, Java etc., because instead of exporting a callback-style iterator via `HASH_ITER`, these macros iterates in dataset-style (top/next) using a `boolean` type, signaling the end of hash table, like showed in the example below:

```c
#include "uthash.h"
#include <stdio.h>
#include <stdbool.h>

struct my_struct {
    int id;
    char name[20];
    UT_hash_handle hh;
};

struct my_struct *users = NULL;

void add_user(int user_id, char *name) {
    struct my_struct *s;
    s = malloc(sizeof(struct my_struct));
    s->id = user_id;
    strcpy(s->name, name);
    HASH_ADD_INT(users, id, s);
}

void delete_all() {
    struct my_struct *current_user, *tmp;

    HASH_ITER(hh, users, current_user, tmp) {
        HASH_DEL(users, current_user);
        free(current_user);
    }
}

int main() {
    struct my_struct *user, *tmp;
    bool users_eof;

    add_user(100, "Syd Barrett");
    add_user(101, "Roger Waters");
    add_user(102, "David Gilmour");
    add_user(103, "Richard Wright");
    add_user(104, "Nick Mason");

    HASH_TOP(hh, users, user, tmp, &users_eof);
    while (!users_eof) {
        printf("%d - %s\n", user->id, user->name);
        HASH_NEXT(hh, users, user, tmp, &users_eof);
    }

    delete_all();
    return 0;
}
```
```bash
$ ./a.out
100 - Syd Barrett
101 - Roger Waters
102 - David Gilmour
103 - Richard Wright
104 - Nick Mason
```
In general, the languages that offer the `for in` support implements their lists in dataset-style, and the programmer must implement two methods: `GetCurrent()`/`MoveToNext()`. Therefore, `HASH_TOP`/`HASH_NEXT` would be useful for them. :-)

What do you think about this feature? (if it was accepted I finally would be able to create a uthash binding for Pascal :-) )

Thank you!

PR #109, opened 2017-03-17
Added macros HASH_TOP and HASH_NEXT for dataset-style iteration.

silvioprog:
Hello,

This PR 109 adds macros `HASH_TOP` and `HASH_NEXT` for dataset-style iteration. See the example below:

```c
#include "uthash.h"
#include <stdio.h>

struct my_struct {
    int id;
    char name[20];
    UT_hash_handle hh;
};

struct my_struct *users = NULL;

void add_user(int user_id, char *name) {
    struct my_struct *s;
    s = malloc(sizeof(struct my_struct));
    s->id = user_id;
    strcpy(s->name, name);
    HASH_ADD_INT(users, id, s);
}

void delete_all() {
    struct my_struct *current_user, *tmp;

    HASH_ITER(hh, users, current_user, tmp) {
        HASH_DEL(users, current_user);
        free(current_user);
    }
}

int main() {
    struct my_struct *user, *tmp;

    add_user(100, "Syd Barrett");
    add_user(101, "Roger Waters");
    add_user(102, "David Gilmour");
    add_user(103, "Richard Wright");
    add_user(104, "Nick Mason");

    HASH_TOP(hh, users, user, tmp);
    while (user) {
        printf("%d - %s\n", user->id, user->name);
        HASH_NEXT(hh, users, user, tmp);
    }

    delete_all();
    return 0;
}
```
```bash
./a.out
100 - Syd Barrett
101 - Roger Waters
102 - David Gilmour
103 - Richard Wright
104 - Nick Mason
```

### DIFF
--- a/src/uthash.h
+++ b/src/uthash.h
@@ -999,6 +999,34 @@ for(((el)=(head)), ((tmp)=DECLTYPE(el)((head!=NULL)?(head)->hh.next:NULL));     
   (el) != NULL; ((el)=(tmp)), ((tmp)=DECLTYPE(el)((tmp!=NULL)?(tmp)->hh.next:NULL)))
 #endif
 
+#ifdef NO_DECLTYPE
+#define HASH_TOP(hh,head,el,tmp)                                                 \
+do {                                                                             \
+  (el) = (head);                                                                 \
+  *(char**)(&(tmp)) = (char*)((head!=NULL)?(head)->hh.next:NULL);                \
+} while (0)
+#else
+#define HASH_TOP(hh,head,el,tmp)                                                 \
+do {                                                                             \
+  (el) = (head);                                                                 \
+  (tmp) = DECLTYPE(el)((head!=NULL)?(head)->hh.next:NULL);                       \
+} while (0)
+#endif
+
+#ifdef NO_DECLTYPE
+#define HASH_NEXT(hh,head,el,tmp)                                                \
+do {                                                                             \
+  (el) = (tmp);                                                                  \
+  *(char**)(&(tmp)) = (char*)((tmp!=NULL)?(tmp)->hh.next:NULL);                  \
+} while (0)
+#else
+#define HASH_NEXT(hh,head,el,tmp)                                                \
+do {                                                                             \
+  (el) = (tmp);                                                                  \
+  (tmp) = DECLTYPE(el)((tmp!=NULL)?(tmp)->hh.next:NULL);                         \
+} while (0)
+#endif
+
 /* obtain a count of items in the hash */
 #define HASH_COUNT(head) HASH_CNT(hh,head)
 #define HASH_CNT(hh,head) ((head != NULL)?((head)->hh.tbl->num_items):0U)


silvioprog:
Oops, sorry mixing PRs dudes. ^^'

This new PR 109 apply all suggested fixes from Arthur in the [108](https://github.com/troydhanson/uthash/pull/108) comments.

I'm going to send the doc/tests ASAP (probably this weekend), but feel free to apply it.

No more mixing in new PRs. :-)

silvioprog:
Hello again Arthur. :-)

Could you apply this PR? Let me know if it need more improvements. :-)

Thank you!

Quuxplusone:
Well, it's still missing tests and documentation.
I was just looking at where the documentation would go when it struck me... *do* we really need these macros? Your original rationale in #108 was intimately related to your idea about a Pascal binding, but that wouldn't really have anything to do with these macros. Couldn't you equally well rewrite

```
    HASH_TOP(hh, users, user, tmp);
    while (user) {
        printf("%d - %s\n", user->id, user->name);
        HASH_NEXT(hh, users, user, tmp);
    }
```

as simply

```
    user = users;
    while (user) {
        struct my_struct *next = user->hh.next;
        printf("%d - %s\n", user->id, user->name);
        user = next;
    }
```

? Like, what's the benefit of these new macros?  It now seems to me that uthash already has idioms for "high-level, deletion-safe foreach" and "low-level iteration", and we don't really need another idiom.

So I'm backing off of my initial assessment of this patch. Seems like it could use more of a rationale.

Do you have a link to the Python binding you're trying to make?

silvioprog:
I totally agree with you. We need to keep uthash smallest possible, so any new unnecessary macro grows it more and more.

I need to do the tests above ASAP sending new feedback ...

> Do you have a link to the Python binding you're trying to make?

Yes. I'm developing a HTTP server/client C library that uses uthash as hash table for strings, representing the headers, server-cookies, form fields, environment variables and more. Today I'm working only on its Pascal binding, but I intend to create the Python binding after first library release.

I forgot a tip, it would be nice to have some page like "who is using uthash". I'm going to publish my library soon, and I would be glad to say that it uses uthash. :-)

Quuxplusone:
>> Do you have a link

> Yes.

What is the link? ;)

silvioprog:
Oh, sorry, I forgot that. Do you have a Gitlab account? It will be a pleasure to add your user name there.

Today the library is in a private repository because I need to write the documentation and some parts of the code still in draft, but I'm working a lot to release it ASAP under MIT here in Github in a public repository.

Quuxplusone:
@silvioprog: I've never used Gitlab before, but I've just created an account there with the same name I have here (Quuxplusone), so you can add me if you want.

silvioprog:
Done.

Now you can open [this link](https://gitlab.com/silvioprog/brookframework). Don't mind the messy, this repository is just a rack, so there is a lot of sketch there. :-)

As soon as an usable code and documentation is give I'll release it in the official repository: https://github.com/silvioprog/brookframework.

Quuxplusone:
So the problem you're trying to solve with HASH_TOP/HASH_NEXT is that you want to add some new function(s) to https://gitlab.com/silvioprog/brookframework/blob/master/bindings/pascal/libbrook.pas to enable natively iterating over a `Pbf_hs` in Pascal, right? What would that new function(s)' signature look like?

I might be overthinking it. If you just want to provide `top` and `next` functions in the Pascal binding, you can already do that, by working with the `->hh.next` pointers directly in your C code, as in my example [above](https://github.com/troydhanson/uthash/pull/109#issuecomment-287595847). (uthash supports that; it's not considered bad practice as far as I'm concerned.)

silvioprog:
Hello again Arthur, thanks for replying and sorry for late answer! :-) (I need to check why Github doesn't notified me)

You got the exactly point of the problem. I was afraid to use `->hh.next` directly because I was not sure if it would change in future uthash versions, but now that you said it is not bad practice, I did a new version of my API using that, however I still have two doubts. First, this is the new functions I've done:

```c
BF_EXTERN struct bf_hs *bf_hs_next(struct bf_hs *hs) __nonnull((1));
...
struct bf_hs *bf_hs_next(struct bf_hs *hs) {
    return NULL != hs ? hs->hh.next : NULL;
}

// example:

    struct bf_hs *hsl = NULL, *hs = NULL, *hs_next = NULL;
    hs = hsl;
    while (hs) {
        hs_next = bf_hs_next(hs);
        printf("%s\n", bf_hs_name(hs));
        hs = hs_next;
    }
```
So my first doubt is: why not using an `if (hs_next)`, will it always be not null?:

```c
    hs = hsl;
    while (hs) {
        hs_next = bf_hs_next(hs);
        if (hs_next) // <-- here
            printf("%s\n", bf_hs_name(hs));
        hs = hs_next;
    }
```
And the next doubt is, why not using just as below (notice I don't know uthash deeply yet)?:

```c
    hs = hsl;
    while (hs) {
        printf("%s\n", bf_hs_name(hs)); // now I don't need an extra `if` inside my loop :-)
        hs = bf_hs_next(hs);
    }
```
The code above seems good, but I'm not sure if it really works properly, because I've not tested it yet.

silvioprog:
Arthur, perfect! :-)

I can add an entry in the documentation warning that `next` must not be used for deleting/freeing itens.

Now this issue was totally solved, I'm going to commit all changes in my library. Thanks a lot! :+1:

PR #117, opened 2017-04-09
remove libut

troydhanson:
libut was my packaging of uthash plus the other headers, plus the undocumented utvector/utmm C sources, so the whole ball of wax could be used with `#include "libut.h"` and linked with `-lut`. We discussed removing libut from uthash when Arthur became maintainer. This pull request does just that. The motivation is to simplify uthash to its core, documented interfaces.


troydhanson:
Yes that's right. Thanks Arthur!

Issue #93, opened 2016-10-28
Investigate whether the table can be left in a "good" state after uthash_fatal


Quuxplusone:
Basically I'd like to either explicitly support (and unit-test) cases like the following, or explicitly document that these cases are not supported.  But if we do the latter, we'll be explicitly "unsupporting" some code that's already out there in the wild (see #92).

```
#include "uthash.h"
#undef uthash_fatal

#undef uthash_malloc
void *uthash_malloc(size_t sz) {
    static int count = 0;
    if (++count == FAIL_ON_THIS_ALLOCATION) return NULL;
    return malloc(sz);
}

struct Table {
    int hashkey;
    UT_hash_handle hh;
};

bool addItemOrFail(struct Table *t, int newValue) {
    struct Table *item = malloc(sizeof *item);
    item->hashkey = 42;
#define uthash_fatal(msg) return false
    HASH_ADD_INT(table, hashkey, item);
#undef uthash_fatal
    return true;
}

int main() {
    struct Table *t = NULL;
    addItemOrFail(t, 41);    // Suppose this fails.
    addItemOrFail(t, 42);    // Is this undefined behavior, or does it Just Work?
    addItemOrFail(t, 43);
    HASH_CLEAR(hh, t);    // Is this undefined behavior, or does it Just Work?
}
```

Quuxplusone:
@tbeu: Bah, I wrote a big long response on https://trac.modelica.org/Modelica/ticket/2097 and then the spam filter rejected it. Long story short, watch *this* GitHub issue for updates on the definedness of your approach. I now think Modelica should simply `abort()` on malloc failure instead of trying to be clever with control flow.

Either way, I'd strongly prefer if you credited yourself in the Modelica changelog, instead of me, to avoid people thinking they should contact me about Modelica issues just because my name appears high up in the changelog. Thanks!

tbeu:
Any news on this?

Quuxplusone:
Sorry, no news is bad news in this case. I did maybe half of the investigation I was planning to do, and the results look very bad so far.

- The good news: `uthash_fatal` is used *only* to recover from `uthash_malloc` failure. Every call to `uthash_malloc` is paired with a conditional call to `uthash_fatal`.

[...]

PR #121, opened 2017-05-18
Enable non-fatal malloc failures

veselov:
Can you please review and comment?
HASH_SELECT still needs to be re-written.
I tried to do negative tests on uthash_fatal, but it's a bit cumbersome, uthash_fatal must stop the current thread, I can break it up into smaller programs, or use threads, which is new. You can see the test code at https://pastebin.com/aHFXUVZX
Thank you.


src/uthash.h
```diff
@@ -177,24 +193,45 @@ do {
...
 #define HASH_MAKE_TABLE(hh,head)                                                 \
 do {                                                                             \
+  int _mem_ok = 1;                                                               \
   (head)->hh.tbl = (UT_hash_table*)uthash_malloc(sizeof(UT_hash_table));         \
-  if (!(head)->hh.tbl) {                                                         \
+  if (!((head)->hh.tbl)) {                                                       \
+    if (HASH_NOMEM_OK) {                                                         \
+      break;                                                                     \
+    }                                                                            \
     uthash_fatal("out of memory");                                               \
```
Quuxplusone: I'm skeptical of the way this patch *avoids* calling `uthash_fatal` in OOM situations. Shouldn't we notify the client *somehow* that we've hit OOM, and then leave it up to the client ('s `uthash_fatal` handler) to say "exit" or "longjmp" or "just keep going"?

src/uthash.h
```diff
@@ -177,24 +193,45 @@ do {
 
 #define HASH_MAKE_TABLE(hh,head)                                                 \
 do {                                                                             \
+  int _mem_ok = 1;                                                               \
   (head)->hh.tbl = (UT_hash_table*)uthash_malloc(sizeof(UT_hash_table));         \
-  if (!(head)->hh.tbl) {                                                         \
+  if (!((head)->hh.tbl)) {                                                       \
```
Quuxplusone: Do you need these extra parentheses to suppress a compiler warning or something, or is this just a merge conflict?

src/uthash.h
```diff
@@ -423,7 +480,7 @@ do {
  */
 #ifdef HASH_DEBUG
 #define HASH_OOPS(...) do { fprintf(stderr,__VA_ARGS__); exit(-1); } while (0)
-#define HASH_FSCK(hh,head)                                                       \
+#define HASH_FSCK(hh,head,fun)                                                   \
```
Quuxplusone: This seems like a good idea but orthogonal to the error-handling stuff. If you rename `fun` to `where` throughout, and then submit this as a separate patch, I'll accept it.

Quuxplusone: Oh, and parenthesize `(where)` every time it's used in this macro, for hygiene. Doesn't matter for correctness, but I'd like to keep a general rule of "hygienic macros all the time no matter what."

Quuxplusone:
For testing that `uthash_fatal` jumps out as expected, maybe you could use `setjmp`/`longjmp`? You'd have to be okay with leaking the hashtable, since it would be left in a bad state... oh, unless your patch is also fixing that?

src/uthash.h
veselov: Must be a merge conflict. I stepped on your "big white space cleanup".

src/uthash.h
```diff
@@ -146,8 +158,12 @@ do {
 do {                                                                             \
   (tbl)->bloom_nbits = HASH_BLOOM;                                               \
   (tbl)->bloom_bv = (uint8_t*)uthash_malloc(HASH_BLOOM_BYTELEN);                 \
-  if (!(tbl)->bloom_bv) {                                                        \
-    uthash_fatal("out of memory");                                               \
+  if (!((tbl)->bloom_bv))  {                                                     \
+    if (HASH_NOMEM_OK) {                                                         \
+      _mem_ok = 0;                                                               \
```
veselov: I take it that is so that one macro never changes/reads args from another macro, right?

src/uthash.h
veselov: Well, that was the whole point. If the client wants to use exit/jmp, they should not define `HASH_NOMEM_OK`, and they will get what they want. `HASH_NOMEM_OK` explicitly turns on the mode where error handling is to be done by checking `hh.mem_failed` after a hash write operation. This should cover all 3 cases, and I don't see a point when you want to mix them together somehow. Doing exit/jmp will prevent `mem_failed` to be set up properly, it will also prevent clean up (you really really don't want to use `longjmp` or `exit_thread` just because of that). Doing anything else but jmp/exit will be questionable because it carries no information on where the failure occurred. You might as well deal with it by checking `mem_failed` after an operation.

Quuxplusone: One concern I have is that "a hash write operation" might not be very well defined, in the mind of the client programmer. The model of "just check `hh.mem_failed` after every hash operation" is *very* uncomfortably close to "just check `errno` after every library operation", which is anathema to me. If the client could set

    #define uthash_kinda_fatal(x) do { log("uh-oh! hash out of memory!"); } while (0)

then I'd feel much better.

src/uthash.h
Quuxplusone: Basically yes. Think about all these macros as if they were C functions; what you're doing here is analogous to using a global variable. (Actually it's worse, because it's a global variable that may or may not exist depending on who called me.)  In C we'd fix that by passing in the address of the variable to modify as an `int*` function parameter; in macro-land we fix it by passing in the *name* of the variable to modify as a macro parameter.

src/uthash.h
veselov: You are saying it's an anathema to check `errno`/return values from every library call, but wait, wouldn't very bad things happen if you didn't? Usually in a chain of functions if one fails, all others wouldn't make sense, and it takes craft to either make the rest be noops, or unwind at the end. Yes, it's very tedious and probably are responsible for a few folks to change their professions. That's why exceptions are so popular, because they let you not check the result, and just deal with any error in a concentrated place.

But, there in the above I convinced myself. People can sure want to use exceptions there if they are using it in C++ code. Even breaking out of their do{}while() loops that so conveniently mimic the exception model (labels required).

How about this : if when `HASH_NOMEM_OK` is defined, `uthash.h` also defines, if not set, `uthash_failed`, with the arguments of `element` (to be filled in with the element that is somehow involved in the failure), to do nothing (the caller has to define it with a meaningful filling). Then, at the end of any potential clean up, and after `mem_failed` is set to 1, the `uthash_failed` is called.

veselov:
@Quuxplusone Can you please re-review?
HASH_SELECT gave me a bit of grief, so I was stuck on it for a while.
Quick recap:

1. There is now a `uthash_mem_failed` macro that gets called if `HASH_NOMEM_OK` is 1 (and `malloc()` fails). This enable callback based behavior.
2. I should have cleaned up all formatting. Please do let me know if I missed anything
3. I had to change `HASH_DELETE` (now `HASH_DELETE_IFBUCKET`), because otherwise it wasn't possible to call from `HASH_SELECT`
4. I didn't refactor `HASH_SELECT` as @troydhanson suggested in #93 , I don't think it's possible to call `HASH_ADD` because the type is unknown.

src/uthash.h
```diff
@@ -142,12 +158,16 @@ do {
...
 #define HASH_BLOOM_BYTELEN (HASH_BLOOM_BITLEN/8UL) + (((HASH_BLOOM_BITLEN%8UL)!=0UL) ? 1UL : 0UL)
-#define HASH_BLOOM_MAKE(tbl)                                                     \
+#define HASH_BLOOM_MAKE(tbl,mem_ok)                                              \
 do {                                                                             \
   (tbl)->bloom_nbits = HASH_BLOOM;                                               \
   (tbl)->bloom_bv = (uint8_t*)uthash_malloc(HASH_BLOOM_BYTELEN);                 \
-  if (!(tbl)->bloom_bv) {                                                        \
-    uthash_fatal("out of memory");                                               \
+  if (!((tbl)->bloom_bv))  {                                                     \
+    if (HASH_NOMEM_OK) {                                                         \
```
Quuxplusone: I'd still like to see this code block defined as a macro somewhere up top, and then used here without any conditional branches. Although, I also don't like the use of `break` to "skip the rest of this macro", because it's not robust to refactoring — once a macro contains a `break`, you can never pull that bit out into a shared subroutine-macro.

In other words, I'm hoping for something like this:

    #if HASH_NOMEM_OK
    #define HASH_OOM(mem_ok)  (mem_ok) = 0
    #else
    #define HASH_OOM(mem_ok) uthash_fatal("out of memory")
    #endif

    if (!(tbl)->bloom_bv) {
        HASH_OOM(mem_ok);
    } else {
        uthash_bzero((tbl)->bloom_bv, HASH_BLOOM_BYTELEN);
        (tbl)->bloom_sig = HASH_BLOOM_SIGNATURE; 
    }

but I know that might be a lot harder to pull off, with complicated things like `HASH_ADD`.

src/uthash.h
```diff
@@ -105,6 +105,22 @@ typedef unsigned char uint8_t;
...
+#define SET_MEM_FAILED(hh,v) (hh)->mem_failed = (v)
+#define GET_MEM_FAILED(hh) ((hh)->mem_failed)
+#else
+#define HASH_NOMEM_OK 0
+#define SET_MEM_FAILED(hh,v) do{}while(0)
+#define GET_MEM_FAILED(hh) 0
+#endif
+
+#ifndef uthash_mem_failed
+#define uthash_mem_failed(elem) do{}while(0)
```
Quuxplusone: I don't understand the intended purpose of `uthash_mem_failed`. (In particular, what does its parameter represent?)

src/uthash.h
```diff
@@ -177,24 +197,45 @@ do {
...
+    }                                                                            \
+    uthash_bzero((head)->hh.tbl->buckets,                                        \
+        HASH_INITIAL_NUM_BUCKETS * sizeof(struct UT_hash_bucket));               \
+    HASH_BLOOM_MAKE((head)->hh.tbl,_mem_ok);                                     \
+    if (HASH_NOMEM_OK && !_mem_ok) {                                             \
+      break;                                                                     \
+    }                                                                            \
+    (head)->hh.tbl->signature = HASH_SIGNATURE;                                  \
+  } while (0);                                                                   \
+  if (HASH_NOMEM_OK && !_mem_ok) {                                               \
```
Quuxplusone: I'm not a fan of this repeated `if` test, for a condition that most of the time will be false, and if it is true, we knew it was true when we jumped here. Can you refactor this to get rid of the test? (Obviously it would be easy with a `goto`, but I think we have to avoid `goto` also. ;))

This is also a case of `if (HASH_NOMEM_OK ...` which should be handled with a `#define HASH_MAKE_TABLE_OOM(...)` somewhere up top.

src/uthash.h
```diff
@@ -323,13 +375,24 @@ do {
     (add)->hh.prev = NULL;                                                       \
     (head) = (add);                                                              \
     HASH_MAKE_TABLE(hh, head);                                                   \
+    if (HASH_NOMEM_OK && !(head)->hh.tbl) {                                      \
+      (head) = 0;                                                                \
```
Quuxplusone: Lines 376–379 could almost certainly be refactored to `HASH_MAKE_TABLE(hh, add)` and then assign `(head) = (add)` only if the `HASH_MAKE_TABLE` succeeded.

src/uthash.h
```diff
@@ -1098,6 +1191,9 @@ typedef struct UT_hash_handle {
    void *key;                        /* ptr to enclosing struct's key  */
    unsigned keylen;                  /* enclosing struct's key len     */
    unsigned hashv;                   /* result of hash-fcn(key)        */
+#if HASH_NOMEM_OK == 1
+   int mem_failed;
+#endif
```
Quuxplusone: I definitely don't think the struct layout of `UT_hash_handle` should change based on `-DHASH_NOMEM_OK`; the name of the macro doesn't *sound* like it would affect the ABI.

Why does this flag need to live in the hash handle? Couldn't it live in `UT_hash_table` instead? This would still be an ABI breakage, but it wouldn't also be increasing the size of the user's structs; it would be `O(1)` extra space instead of `O(n)` extra space.  And actually `UT_hash_table` has plenty of extra padding bits on 64-bit platforms and/or when `HASH_BLOOM` is defined, some of which could be filled with a `uint8_t mem_failed` flag.

src/uthash.h
```diff
@@ -323,13 +375,24 @@ do {
     (add)->hh.prev = NULL;                                                       \
     (head) = (add);                                                              \
     HASH_MAKE_TABLE(hh, head);                                                   \
+    if (HASH_NOMEM_OK && !(head)->hh.tbl) {                                      \
+      (head) = 0;                                                                \
+      SET_MEM_FAILED(&(add)->hh, 1);                                             \
+      uthash_mem_failed(add);                                                    \
```
Quuxplusone: FWIW, I still haven't figured out the logic by which sometimes we do `mem_ok = 0`, sometimes we do `SET_MEM_FAILED`, sometimes we do `uthash_mem_failed`, and sometimes we do `uthash_fatal("out of memory")`. I suspect we might be able to just get rid of `uthash_mem_failed`; but are the other 2 or 3 mechanisms really irreducible?

src/uthash.h
```diff
@@ -105,6 +105,22 @@ typedef unsigned char uint8_t;
 #define uthash_expand_fyi(tbl)            /* can be defined to log expands   */
 #endif
 
+/* whether we can recover from malloc() failure */
+#ifdef HASH_NOMEM_OK
```
veselov: Will change. My idea was - if you define HASH_NOMEM_OK (to anything), then we turn the behavior on. There was no promise that setting HASH_NOMEM_OK to 0 would "disable" the behavior somehow. The 0/1 value of the macro was intended for internal purposes only.

src/uthash.h
veselov: This is the callback for memory failures. I thought you were adamant that it should exist (so the caller doesn't check the result mem_ok field). The parameter is the element that failed to add (or involved in memory failure in some way, but ATM, memory is allocated only when elements are being added).

src/uthash.h
veselov: `mem_ok` (local var) is used internally as a parameter for macros that don't have access to hh, e.g. `HASH_BLOOM_MAKE`. I can get rid of `mem_ok` by altering those macros and adding the hh as a parameter to those macros instead. Then, the `hh.mem_failed` will be used instead. OK?

`uthash_mem_failed()` is the callback mechanism. You were adamant that one is needed. I agreed that is useful, because folks may use exceptions, or even `longjmp()` (the memory state will not remain corrupt unlike if used from `uthash_fatal`) as a simpler mechanism of dealing with OOMs. Use of `uthash_mem_failed` and `hh.mem_failed` is complementary to each other. If one had to go on the chopping block, however, it will have to be `uthash_mem_failed()`, because there is no universal access to exceptions, or use of `longjmp()`, however checking 'hh.mem_failed` is always available.

src/uthash.h
veselov: Is it OK if we simply state that if `elem.hh.tbl` is 0 after an add operation, there was a memory failure? That sounds consistent to me.

veselov:
@Quuxplusone To summarize the review exchanges, can you please comment on those:
1. `uthash_mem_failed()` macro stands, is invoked as last operation for an OOM, only for `NOMEM_HASH_OK`
1. Use of `break` and repeated checks is to be eliminated. However, this will lead to a lot more `if`s, nesting and repeated recovery calls.
1. Local variables are to be used for carrying allocation failures. `UT_hash_handle`, however, will have no flags, and the user can check that allocation failed by checking `hh.tbl` for `NULL`. 

Rest of your review comments will be addressed.

Quuxplusone:
1. IIUC, `uthash_mem_failed()` is basically the "non-fatal" variant of `uthash_fatal()`? Since `uthash_fatal()` right now is only ever called for OOM conditions, I think it would be simplest to just accept that `uthash_fatal` is poorly named, and treat it as if it were called `uthash_oom` instead — that is, we should call `uthash_fatal` after detecting OOM, no matter whether `HASH_NOMEM_OK` is true or false. Just, if `HASH_NOMEM_OK` is true, we should expect that `uthash_fatal` will return instead of exiting/longjmping.

2. Use of `break` is at least frowned upon. I can't say for sure that it's going to be a good idea to get rid of it. Repeated checks should be eliminated; and in particular, when `HASH_NOMEM_OK` is false, there should be no extra `if`s in the C code — not even `if (false)`s.

3. I think that checking `hh.tbl` for `NULL` sounds fine. (That might even be wrapped into a macro spelled something like `HASH_ENCOUNTERED_OOM(head)`.) I worry that it won't always be obvious to the user which `head` needs to be checked. But, if the user really wants to know when they hit OOM, they can override `uthash_fatal` (formerly known as `uthash_mem_failed`) to tell them for sure.

I envision the "safe" idiom being something like this:

```
#define HASH_NOMEM_OK  // or HASH_NONFATAL or whatever: bikeshed
#include "uthash.h"
#undef uthash_fatal
...
    for (int i=0; i < NELEM(usertab); ++i) {
        bool hit_oom = false;
#define uthash_fatal(msg) hit_oom = true
        HASH_ADD_INT(users,id,usertab[i]);
#undef uthash_fatal
        if (hit_oom) {
            log("Inserted %u users before OOM", HASH_COUNT(users));
            HASH_CLEAR(users);
            return -1;
        }
    }
```

veselov:
1. `uthash_mem_failed()` is like `uthash_fatal()`, right. However, it also has an argument that points to the element that is involved. I'd rather preserved that, because it enables setting flags inside the element itself, if the wants to. This, IMHO, can be quite helpful. It's especially helpful in something like `HASH_SELECT`, where multiple elements may fail to insert (but in a stream of elements, for example, first may fail to insert, but the following may all succeed). Also, the code will work fine if `uthash_mem_failed()` did not return (e.g. threw or similar), this still won't leave any indeterminate state around. So, if you are OK with the above, I'll re-use the same name, that's not a problem. However, the default will change (from `exit` to a no-op), and the signature will be different; the locations at which the calls will be made will also change, obviously.

2. All right, let's see what it'll look like first

3. I'm not sure how looking at the head is helpful in this case. AFAIU, you can do:
```
#define HASH_NOMEM_OK  // or HASH_NONFATAL or whatever: bikeshed
#include "uthash.h"
...
    for (int i=0; i < NELEM(usertab); ++i) {
        HASH_ADD_INT(users,id,usertab[i]);
        if (!usertab[i].hh.tbl) {
            log("Inserted %u users before OOM", HASH_COUNT(users));
            HASH_CLEAR(users);
            return -1;
        }
    }
```

src/uthash.h
```diff
@@ -984,15 +1063,29 @@ do {
...
+            }                                                                    \
           } else {                                                               \
             _dst_hh->tbl = (dst)->hh_dst.tbl;                                    \
           }                                                                      \
-          HASH_TO_BKT(_dst_hh->hashv, _dst_hh->tbl->num_buckets, _dst_bkt);      \
-          HASH_ADD_TO_BKT(_dst_hh->tbl->buckets[_dst_bkt], _dst_hh);             \
+          if (!HASH_NOMEM_OK || (_dst_hh->tbl != NULL)) {                        \
+            HASH_TO_BKT(_dst_hh->hashv, _dst_hh->tbl->num_buckets, _dst_bkt);    \
+            HASH_ADD_TO_BKT(_dst_hh->tbl->buckets[_dst_bkt], hh_dst, _dst_hh);   \
           HASH_BLOOM_ADD(_dst_hh->tbl, _dst_hh->hashv);                          \
```
Quuxplusone: nit: merge conflict here

ziab:
Came across because ran into the same problem, my program needs to handle mem allocation issues gracefully. BTW, thanks for working on this issue, handling memory allocation issues is very important feature. 
So, I took the change created by @veselov, my low-resource simulation test passes now. Just wanted to confirm if I'm using this correctly, given this code snippet:

		        struct hash_entry
		        {
			    t_key _key;
			    t_val _val;
			    UT_hash_handle hh;  /* makes this structure hashable */
		        };

			hash_entry* new_entry = new hash_entry;
			if (!new_entry) return false;

			new_entry->_key = key;
			new_entry->_val = val;

			HASH_ADD_INT(hash_table, _key, new_entry);

			if (hash_table && hash_table->hh.mem_failed == 1 || new_entry->hh.mem_failed == 1)
			{
				// mem allocation failure occured
				return false;
			}

			return true;

is this a correct way to check failure: `if (hash_table && hash_table ->hh.mem_failed == 1 || new_entry->hh.mem_failed == 1)` i.e. I need to check both the `hash_table` (because if it's the first insert, the hash_table might not be initialized yet), the `new_entry` (because it might fail to be inserted). 
Is that right?

Perhaps I should declare similar macro to `#define uthash_mem_failed(e) do{((example_user_t*)e)->mem_failed=1;}while(0)` as suggested by one of the unit test and have my `hash_entry` declare a `mem_failed` data member. What would be the difference in this case?

Sorry if I'm asking something that has been answered before, I'm probably missed something.

veselov:
@ziab 
1. You only need to check whether `new_entry->hh.mem_failed == 1`
2. There would be no difference if you did it through `uthash_mem_failed`, except that:
- you may have to clear the corresponding flag yourself before the operation
- the mem_failed macro is more intended to be used with something like a `throw`/`setjmp()`, the way it is used in the test case is just for testing that it works.
3. Note that the API will change in the next patch

ziab:
@veselov thanks for clarification! Looking forward to the final version.

veselov:
@Quuxplusone I'm trying to refactor `HASH_ADD_KEYPTR_BYHASHVALUE_INORDER` and `HASH_ADD_KEYPTR_BYHASHVALUE`. I'm having a bit of an issue refactoring them without either a break or a double check. So, any suggestions will be appreciated.

In essence, the current code does:
```
do {
  if () {
     A
  } else {
     B
  }
  C
} while(0);
```

`A` and `C` can incur OOM. It's trivial to deal with for `C`. But for `A`, the options are not that simple.
1. Break out from `A` - that's what you can see in this branch
1. I can wrap everything into another `do`/`while` loop, and break out of that. I'm not sure if that is any better, I don't think I understand your concerns about using breaks.
1. Duplicate `C` into first `if`/`else` branch, and remove it from where it is:
```
do {
  if () {
     A
     C
  } else {
     B
     C
  }
} while(0);
```
There is a lot of stuff in `C` for this to be reasonable as well.

Quuxplusone:
@veselov: I haven't looked at the actual code again since 6 days ago, but going off of your comment... I think I like the sound of your option number 3 ("Do a second check before C - e.g. A sets a flag, and we don't execute C if so") — with the wrinkle that you should put the "check" and the "sets a flag" both under macros that `#define` away to nothing if `HASH_NOMEM_OK` is not defined.  Or else your option number 4 ("Duplicate C"), *ideally* with some macro that removes the duplication if `HASH_NOMEM_OK` is not defined, but I admit I don't intuitively see how to do that.

veselov:
@Quuxplusone All right, how does that look?
Note that calls to `uthash_fatal()` in case of "nomem" are later, they are done after any and all clean up (so that the caller may throw).
This can be simplified a bit more, if both cases (safe alloc or not) clean up after themselves. But this will change existing behavior (without breaking any compatibility, though).

veselov:
@Quuxplusone bump?

src/uthash.h
```diff
@@ -105,6 +102,30 @@ typedef unsigned char uint8_t;
...
 
+#ifndef HASH_NOMEM_OK
+#define HASH_NOMEM_OK 0
+#endif
+
+#if HASH_NOMEM_OK
+/* malloc failures can be recovered from */
+
+#ifndef uthash_fatal
+#define uthash_fatal(obj) do{}while(0)
```
Quuxplusone: It looks like you're overloading the name `uthash_fatal` to mean two very different things:

- If `!HASH_NOMEM_OK`, it should terminate the program, takes a string literal, and is called on every OOM.
- If `HASH_NOMEM_OK`, it need not terminate the program, takes a hash element, and is called only on OOMs that happen inside `HASH_SELECT`.

In the second case, `uthash_fatal` is the wrong name because it's *not* called under unrecoverable circumstances and it's *not* required to be "fatal" to the program. But `uthash_oom` would also be a wrong name, so long as we're not calling it on every OOM. Now, I think you *should* call it on every OOM — that would suitably fix that problem, and then we could plausibly name it `uthash_oom(obj, where)`.

I think `uthash_oom` should take `(obj, where)` for the same reason that `HASH_FSCK` takes `where`: it might help the programmer to debug the issue.

src/uthash.h
```diff
@@ -983,14 +1058,26 @@ do {
           }                                                                      \
           if ((dst) == NULL) {                                                   \
             DECLTYPE_ASSIGN(dst, _elt);                                          \
-            HASH_MAKE_TABLE(hh_dst, dst);                                        \
+            HASH_MAKE_TABLE(hh_dst, dst, _mem_ok);                               \
+            if (HASH_NOMEM_OK && !_mem_ok) {                                     \
+              uthash_fatal(_elt);                                                \
+              dst = NULL;                                                        \
+              continue;                                                          \
+            }                                                                    \
```
Quuxplusone: One thing you can/should do here to simplify the code is to introduce a macro at the top of uthash.h like this:

    #define IF_HASH_NOMEM_OK(x) x
    #define IF_HASH_NOMEM_OK(x)

and then down here,

    HASH_MAKE_TABLE(hh_dst, dst, _mem_ok);
    IF_HASH_NOMEM_OK(
      if (!_mem_ok) {
        uthash_fatal(_elt);
        dst = NULL;
        continue;
      }
    )

src/uthash.h
```diff
@@ -983,14 +1058,26 @@ do {
...
           } else {                                                               \
             _dst_hh->tbl = (dst)->hh_dst.tbl;                                    \
           }                                                                      \
           HASH_TO_BKT(_dst_hh->hashv, _dst_hh->tbl->num_buckets, _dst_bkt);      \
-          HASH_ADD_TO_BKT(_dst_hh->tbl->buckets[_dst_bkt], _dst_hh);             \
-          HASH_BLOOM_ADD(_dst_hh->tbl, _dst_hh->hashv);                          \
+          HASH_ADD_TO_BKT(_dst_hh->tbl->buckets[_dst_bkt], hh_dst, _dst_hh,      \
+              _mem_ok);                                                          \
```
Quuxplusone: Style nit: I'd prefer you not wrap this line; just go ahead and let it extend past the usual margin.

    HASH_TO_BKT(_dst_hh->hashv, _dst_hh->tbl->num_buckets, _dst_bkt);      \
    HASH_ADD_TO_BKT(_dst_hh->tbl->buckets[_dst_bkt], hh_dst, _dst_hh, _mem_ok); \
    (dst)->hh_dst.tbl->num_items++;                                        \

src/uthash.h
```diff
@@ -264,17 +295,59 @@ do {
 } while (0)
 #endif
 
-#define HASH_ADD_KEYPTR_BYHASHVALUE_INORDER(hh,head,keyptr,keylen_in,hashval,add,cmpfcn) \
+#if HASH_NOMEM_OK
+
+#define HASH_ADD_TO_TABLE(hh,head,keyptr,keylen_in,hashval,add,op,mem_ok)        \
```
Quuxplusone: I think I like this. :)
Nit: `op` should be spelled `where`, for consistency with `HASH_FSCK`. But on line 319 I'm proposing you hoist the `HASH_FSCK` out of this function, at which point you don't need `op`/`where` as a parameter at all, I think.

src/uthash.h
```diff
@@ -264,17 +295,59 @@ do {
...
+      uthash_mem_failed(add);                                                    \
+    } else {                                                                     \
+      HASH_BLOOM_ADD((head)->hh.tbl, hashval);                                   \
+      HASH_EMIT_KEY(hh, head, keyptr, keylen_in);                                \
+    }                                                                            \
+  } else {                                                                       \
+    (add)->hh.tbl = NULL;                                                        \
+    uthash_mem_failed(add);                                                      \
+  }                                                                              \
+  HASH_FSCK(hh, head, op);                                                       \
```
Quuxplusone: Nit: This `HASH_FSCK` should probably be hoisted out into the callers of `HASH_ADD_TO_TABLE`, to avoid repeating it on lines 319 and 332. I also wonder whether avoiding the expensive `HASH_FSCK` in here would allow you to use some part of `HASH_ADD_TO_TABLE` in the implementation of `HASH_SELECT`...?

src/uthash.h
```diff
@@ -365,37 +432,40 @@ do {
...
     uthash_free((head)->hh.tbl->buckets,                                         \
                 (head)->hh.tbl->num_buckets * sizeof(struct UT_hash_bucket));    \
     uthash_free((head)->hh.tbl, sizeof(UT_hash_table));                          \
     (head) = NULL;                                                               \
   } else {                                                                       \
     unsigned _hd_bkt;                                                            \
-    _hd_hh_del = &((delptr)->hh);                                                \
-    if ((delptr) == ELMT_FROM_HH((head)->hh.tbl, (head)->hh.tbl->tail)) {        \
-      (head)->hh.tbl->tail = HH_FROM_ELMT((head)->hh.tbl, (delptr)->hh.prev);    \
+    _hd_hh_del = delhh;                                                          \
```
Quuxplusone: Hygiene nit: `(delhh)`.

src/uthash.h
```diff
@@ -825,44 +898,45 @@ do {
...
-        _he_newbkt->expand_mult = _he_newbkt->count / tbl->ideal_chain_maxlen;   \
-      }                                                                          \
-      _he_thh->hh_prev = NULL;                                                   \
-      _he_thh->hh_next = _he_newbkt->hh_head;                                    \
-      if (_he_newbkt->hh_head != NULL) {                                         \
-        _he_newbkt->hh_head->hh_prev = _he_thh;                                  \
+    HASH_MEM_FAILED(mem_ok);                                                     \
+  } else {                                                                       \
+    uthash_bzero(_he_new_buckets,                                                \
+        2UL * tbl->num_buckets * sizeof(struct UT_hash_bucket));                 \
```
Quuxplusone: Hygiene nit: This should really be `(tbl)` throughout. (If you submit this as a separate PR, I'll merge it.)

src/uthash.h
veselov: Previously I had them separated into `uthash_fatal` and `uthash_mem_failed`, and you have requested that they should be joined together because they are both called in case of OOM, and are nearly identical (and poorly named, yes, but that should be accepted). You are now requesting that they are to be separated. Can you please confirm?

The `uthash_oom` is to be called for every OOM (I'm talking for `HASH_NOMEM_OK=1` case only here). Granted, it is not called at the exact time of the OOM, but only after we released any memory that has been allocated, and restored all state to before the hash call was attempted;  all that is because we allow the caller to jump out of the current stack (using `throw`/`longjmp`, etc.). If an OOM has occurred during the add process, no more memory allocation attempts are performed, so you also can't have multiple malloc failures per call, making it called once per OOM. Even in case of `HASH_SELECT`, it's called once per OOM.

veselov: The `where` parameter you are suggesting, should it contain the same value as `where` for `HASH_FSCK`, or indicate which one of the `malloc`s failed?

Quuxplusone: `where`: I think it should be a string literal; and beyond that, if we can provide more information, we probably should.
Having it be the same as what's passed to `HASH_FSCK` is certainly fine with me.

Quuxplusone: > The `uthash_oom` [you mean the thing called `uthash_fatal` in the current patch, right?] is to be called for every OOM (I'm talking for HASH_NOMEM_OK=1 case only here).

Can you walk me through how it would get called, in the `HASH_NOMEM_OK=1` case, if the user called `HASH_ADD_KEYPTR_BYHASHVALUE_INORDER` and the failure happened, let's say, while allocating the initial bloom filter in `HASH_BLOOM_MAKE`?  I see `HASH_BLOOM_MAKE` calls `HASH_MEM_FAILED` to set `mem_ok=0`, but I don't see anyone calling `uthash_fatal` based on that.

veselov: > The uthash_oom [you mean the thing called uthash_fatal in the current patch, right?] ...

Yes
> Can you walk me through how it would get called, in the HASH_NOMEM_OK=1 case

First, there is my bad; I forgot to rename `hash_mem_failed` back to `hash_fatal`, that's probably what's caused the confusion. And the test didn't catch that because it used the old name too. This PR has been updated with the corresponding changes.

The calls are in `HASH_SELECT` and `HASH_ADD_TO_TABLE`. There are invoked at the exit points of the operations, if it is apparent (`mem_ok` is 0) that adding the element failed.

> if the user called `HASH_ADD_KEYPTR_BYHASHVALUE_INORDER` and the failure happened, let's say, while allocating the initial bloom filter in `HASH_BLOOM_MAKE`?

For bloom allocation to fail, this must be a new table. So, `HASH_MAKE_TABLE` will be called, when bloom is attempted allocation, and failed, _mem_ok will be reset to 0 (provided by `...INORDER`), buckets and table will be freed, and execution will return to `...INORDER`.

The list operations are skipped (because the table is new), and `HASH_ADD_TO_TABLE` is called. The `_mem_ok` is passed there. The `HASH_ADD_TO_TABLE` only executes the actual operations if passed in `mem_ok` is 1, otherwise the element has its table value set to 0, and `uthash_fatal` called on that element. The hash is checked, and `HASH_ADD_TO_TABLE` returns. Note that the same will happen (element's table reset to 0, and `uthash_fatal` called) if the bucket extension fails, this is handled in the other branch of the top-level `if` of `HASH_ADD_TO_TABLE`.

veselov:
@Quuxplusone Here is another version. It addresses all of your previous comments except for providing reason for `uthash_fatal`. Here is why - If we actually do say which operation has failed - since `uthash_fatal()` is invoked after clean up, we have to track the operation down from where the failure actually happened. This means turning `_mem_ok` into a pointer and all that. Even then, most of the time it's probably going to be rehashing. If we were to provide which operation was running when malloc failed, then it's almost always going to be `HASH_ADD_KEYPTR_BYHASHVALUE_INORDER`, unless you are doing something advanced. Please let me know if you still want it, and which should it be (source macro, or indicator what the memory was for).

src/uthash.h
```diff
@@ -365,37 +432,40 @@ do {
...
     }                                                                            \
     if (_hd_hh_del->next != NULL) {                                              \
       HH_FROM_ELMT((head)->hh.tbl, _hd_hh_del->next)->prev = _hd_hh_del->prev;   \
     }                                                                            \
-    HASH_TO_BKT(_hd_hh_del->hashv, (head)->hh.tbl->num_buckets, _hd_bkt);        \
-    HASH_DEL_IN_BKT(hh, (head)->hh.tbl->buckets[_hd_bkt], _hd_hh_del);           \
+    if (from_bucket) {                                                           \
+      HASH_TO_BKT(_hd_hh_del->hashv, (head)->hh.tbl->num_buckets, _hd_bkt);      \
+      HASH_DEL_IN_BKT(hh, (head)->hh.tbl->buckets[_hd_bkt], _hd_hh_del);         \
+    }                                                                            \
```
Quuxplusone: You can probably guess I'm not thrilled with this construct, either. I see that `from_bucket` is always `0` or `1`, so this expands to either `if (0)` or `if (1)`.

What if you just left this code alone (functionally speaking; I do like your refactors above and will probably commit them to master soon) and let it run `HASH_TO_BKT` and a no-op `HASH_DEL_IN_BKT` in the failure case? If `_hd_hh_del->hh_prev` and `->hh_next` are always `NULL`, then this will be fine — and we don't really need to optimize the OOM case for speed, right?

src/uthash.h
```diff
@@ -142,15 +165,16 @@ do {
...
 #define HASH_BLOOM_BITLEN (1UL << HASH_BLOOM)
 #define HASH_BLOOM_BYTELEN (HASH_BLOOM_BITLEN/8UL) + (((HASH_BLOOM_BITLEN%8UL)!=0UL) ? 1UL : 0UL)
-#define HASH_BLOOM_MAKE(tbl)                                                     \
+#define HASH_BLOOM_MAKE(tbl,mem_ok)                                              \
 do {                                                                             \
   (tbl)->bloom_nbits = HASH_BLOOM;                                               \
   (tbl)->bloom_bv = (uint8_t*)uthash_malloc(HASH_BLOOM_BYTELEN);                 \
-  if (!(tbl)->bloom_bv) {                                                        \
-    uthash_fatal("out of memory");                                               \
+  if (!((tbl)->bloom_bv))  {                                                     \
```
Quuxplusone: Style nit: lose one pair of parens here

doc/userguide.txt
```diff
@@ -1595,8 +1595,22 @@ be modified by re-defining the `uthash_fatal` macro.
...
 The fatal function should terminate the process or `longjmp` back to a safe
-place. Uthash does not support "returning a failure" if memory cannot be
-allocated.
+place. Note that an allocation failure may leave allocated memory that can not
+be recovered, and the entire hash table should be considered unusable.
+
+To enable supporting "returning a failure" if memory cannot be
+allocated, define HASH_NOMEM_OK to `1` before including the `uthash.h`
+header file. In this case, the `uthash_fatal` macro defaults to a no-op, any
+memory that is allocated incompletely is cleaned up.
```
Quuxplusone: ...is cleaned up *before invoking the `uthash_fatal` macro.*

tests/test93.c
```diff
@@ -0,0 +1,123 @@
...
+    users = 0; \
+    user = (example_user_t*)malloc(sizeof(example_user_t)); \
+    user->id = id; \
+    is_fatal = 0; \
+    malloc_cnt = a; \
+    /* printf("adding to hash...\n"); */ \
+    if (!setjmp(j_buf)) { \
+        HASH_ADD_INT(users, id, user); \
+    } \
+} while(0)
```
Quuxplusone: Nit: This doesn't *need* to be a macro, does it? Could you make it a regular function?

tests/test93.c
```diff
@@ -0,0 +1,123 @@
...
+        return 1;
+    }
+
+    // let's add users until expansion fails.
+    users = 0;
+    malloc_cnt = 4;
+    while (1) {
+        user = (example_user_t*)malloc(sizeof(example_user_t));
+        user->id = id;
+        if (id++ == 1000) {
```
Quuxplusone: Could this be expressed as a `for` loop?

veselov: It can be, but without an exit condition, so it really would be `for(;id++;)`, so I'm not sure there is any benefit...

src/uthash.h
```diff
@@ -168,33 +192,42 @@ do {
...
+      uthash_bzero((head)->hh.tbl->buckets,                                      \
+          HASH_INITIAL_NUM_BUCKETS * sizeof(struct UT_hash_bucket));             \
+      HASH_BLOOM_MAKE((head)->hh.tbl, mem_ok);                                   \
+      if (HASH_NOMEM_OK && !(mem_ok)) {                                          \
+        uthash_free((head)->hh.tbl->buckets,                                     \
+            HASH_INITIAL_NUM_BUCKETS*sizeof(struct UT_hash_bucket));             \
+        uthash_free((head)->hh.tbl, sizeof(UT_hash_table));                      \
+      } else {                                                                   \
+        (head)->hh.tbl->signature = HASH_SIGNATURE;                              \
+      }                                                                          \
```
veselov: This (control flow) was exactly the reason I didn't do this before. I figured the `HASH_SIGNATURE` assignment was deliberately done at the end to make sure the hash didn't complete initialization, but had that field set. Anyway, fixed.

src/uthash.h
veselov: @Quuxplusone OK, but I have clear out the `hh_prev` and `hh_next` along with increasing the item count, or else. See if you're comfortable with that.

tests/test93.c
```diff
@@ -0,0 +1,123 @@
...
+        HASH_ADD_INT(users, id, user); \
+    } \
+} while(0)
+
+    // $TODO: this test doesn't work because
+    // uthash_fatal can't successfully return - it
+    // makes uthash write into *0. This either needs to
+    // be broken up into small test files, or use threads,
+    // a thread can hard exit. But there are no threads 
+    // in other tests, which means compilation changes, etc.
```
veselov: The comment is no longer valid, I did remove it. By`write into *0` I meant that uthash would attempt writing data at address `0`, causing `SIGSEGV`

veselov:
@Quuxplusone OK, all of your comments are incorporated, except for `for` loop in `test93.c`. Also, the change to get rid of conditional is somewhat ugly, but only affects OOM flow.

veselov:
ping @Quuxplusone

src/uthash.h
```diff
@@ -115,6 +138,16 @@ typedef unsigned char uint8_t;
 /* calculate the hash handle from element address elp */
 #define HH_FROM_ELMT(tbl,elp) ((UT_hash_handle *)(((char*)(elp)) + ((tbl)->hho)))
 
+#define HASH_ROLLBACK_BKT(hh, head, itemptrhh)                                   \
+do {                                                                             \
+  struct UT_hash_handle *_hd_hh_item = (itemptrhh);                              \
+  unsigned _hd_bkt;                                                              \
+  HASH_TO_BKT(_hd_hh_item->hashv, (head)->hh.tbl->num_buckets, _hd_bkt);         \
+  (head)->hh.tbl->buckets[_hd_bkt].count++;                                      \
```
Quuxplusone: Shouldn't this be `count--`, not `count++`?  Please add a unit-test that would have caught this bug (assuming it is a bug), before fixing it.
Also style nit: the backslashes `\` on lines 147-148 should be aligned.

veselov: @Quuxplusone `count++` is correct. `HASH_DELETE_HH` decreases the count, but if bucket resize fails, the count was **not** increased. So it needs to be increased when rolling back to remain unchanged.

veselov:
@Quuxplusone backslashes aligned, no more pending changes.

veselov:
ping @Quuxplusone

veselov:
@Quuxplusone please? :)

Quuxplusone:
I see it, sorry... just haven't had the time to break out my fine-toothed-comb yet. This is good news in that it means I *don't* see anything obviously/critically wrong with it anymore. ;)  In the meantime, I wonder if @ziab has played with this version yet...

ziab:
I haven't played with the latest version yet, but only with the one dated on ~June 8. That version has been stable, all my stress tests, functionality tests and low-resource simulation passed. The next week, I might be able to pull the latest code and integrate it into my project then run the tests.

Quuxplusone:
@veselov: Okay, I finally took a real look at the patch this weekend. I made some relatively trivial changes and put the resulting patch in #130; please take a look. I think your control flow is good; I didn't make any changes there. I did some major renamings, basically to get rid of the term "nomem" in favor of "oom":

- `HASH_NOMEM_OK` -> `HASH_OOM_OK`
- `IF_HASH_NOMEM_OK` -> `IF_HASH_OOM_OK`
- `uthash_fatal` -> `uthash_oom` (only when `HASH_OOM_OK`)
- `HASH_MEM_FAILED(mem_ok)` -> `HASH_RECORD_OOM(oomed)` (and invert the sense of the flag variable)

I rearranged the documentation a little bit (also to account for the fact that `uthash_oom` and `uthash_fatal` are now split up) and made a bunch of mostly cosmetic changes to `test92.c` that also get it working when `HASH_BLOOM` is *not* defined.

I'm still a little bit unhappy with the name `HASH_OOM_OK`since OOM is never exactly "okay", it's just that we're providing a handler for it. But I couldn't think of anything better that mightn't be mistaken for the name of the OOM handler itself. That is, I don't want anyone accidentally writing `-DHASH_OOM_OK=(my handler)`. And vice versa, I thought about just switching the behavior based on the presence of `-Duthash_oom=(my handler)`, but then people would have no way to get the "default" behavior (and `-Duthash_oom` would implicitly define it to `1`, which would just be a syntax error). The naming bikeshed will remain open for about a week at #130, and then I'll merge that one if it LGTY.

Quuxplusone:
Merged 01e5a878d7660e07177a7cb056c769297374d9e3 — THANKS! 🙂

PR #123, opened 2017-06-05
Fixed missing bloom add in HASH_SELECT

veselov:
Apparently HASH_SELECT doesn't add to the bloom filter (if used), making it so nothing can be found.


Quuxplusone:
Merged with a slightly different test case in f3ab1a5426838536a16d912c24f2250ba31b7521 — thanks! `HASH_SELECT` is clearly undertested.

veselov:
@Quuxplusone I swear I made this comment earlier, but I can't understand where it went. `test36.c` doesn't define HASH_BLOOM, which disables use of the bloom filter. Which won't showcase the problem, or am I missing something? That was the reason I added a separate test case, as defining HASH_BLOOM has a side effect

Quuxplusone:
@veselov: You can run the test suite with `EXTRA_CFLAGS=-DHASH_BLOOM=16` to reproduce the bug even with test36.c. (Of course, doing that breaks test6.c; it also breaks with some `-DHASH_FUNCTION=` options. I consider test6's fragility a problem, but at least it's a *separate* problem.)

I'm working on a patch this morning to fix the fragility once and for all, and add a `make thorough` that will run all plausible combinations of `-DHASH_BLOOM`, `-pedantic`, `-DHASH_DEBUG`, `cplusplus`, `-DNO_DECLTYPE`, etc. etc.

Quuxplusone:
@veselov: I'll push [this](https://github.com/Quuxplusone/uthash/commit/4d197f24af391118a7de8dedc2dd272ce32adf9e) later tonight, if it LGTY.

PR #124, opened 2017-06-13
Don't use reserved (C++11) keyword

veselov:
My make thorough fails on nullptr.


Quuxplusone:
Awesome. Not sure how it built on my machine... Thanks!

PR #125, opened 2017-06-15
Shield macro argument for hygiene

veselov:
I have noticed that in case of `fun(macro_arg)`, the parentheses are not applied, but in case of `lval = macro_arg`, they are. I'm wondering what expression will break the latter but not the former (if parentheses were not applied).

### DIFF
--- a/src/uthash.h
+++ b/src/uthash.h
@@ -823,25 +823,25 @@ do {                                                                            
   struct UT_hash_handle *_he_thh, *_he_hh_nxt;                                   \
   UT_hash_bucket *_he_new_buckets, *_he_newbkt;                                  \
   _he_new_buckets = (UT_hash_bucket*)uthash_malloc(                              \
-           2UL * tbl->num_buckets * sizeof(struct UT_hash_bucket));              \
+           2UL * (tbl)->num_buckets * sizeof(struct UT_hash_bucket));            \
   if (!_he_new_buckets) {                                                        \
     uthash_fatal("out of memory");                                               \
   }                                                                              \
   uthash_bzero(_he_new_buckets,                                                  \
-          2UL * tbl->num_buckets * sizeof(struct UT_hash_bucket));               \
-  tbl->ideal_chain_maxlen =                                                      \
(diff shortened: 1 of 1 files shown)


Quuxplusone:
> I have noticed that in case of fun(macro_arg), the parentheses are not applied, but in case of lval = macro_arg, they are. I'm wondering what expression will break the latter but not the former (if parentheses were not applied).

FYI, `a=b` will break the latter but not the former.
The only thing that could break `fun(arg)` is if `arg` contained an unparenthesized comma; but it can't contain commas because it came into this macro as a single argument, and single macro arguments can never contain unparenthesized commas.

veselov:
You mean the gcc warning that requests parens around an assignment?

Quuxplusone:
No, I mean

    a=b = c
    (a=b) = c

have different effects. Or to make it even more obvious,

    (a+=b) = 1;
    a += b = 1;

The former leaves `a==1`; the latter leaves `b==1` and increments `a`.
https://wandbox.org/permlink/QAbFO0yi5AZiQPzC

veselov:
I totally understand the impact of wrapping lval into parens.
But I was only talking about the macro argument is on the right side of the equality sign. I.e. what can `X` be so that `a = X;` is different than `a = (X);`

Quuxplusone:
Ah, sorry, I misinterpreted the question. Talking about the right-hand side, I'm 99% sure you're correct, but it's quite an arcane question. I'm sure it's better and safer to just add the parentheses than constantly worry that we might have missed something. ;)

The "house style" for uthash is that macro arguments should *always* be parenthesized unless

- they are being passed along unchanged in another macro- or function-argument-list (which is safe because of the above comment about the impossibility of their containing unshielded commas), or
- they are callbacks that might themselves be macros or C++ overloaded function names, e.g. the comparison callback to `HASH_SORT` (in which case parenthesizing `(f)(args...)` would be wrong, but parenthesizing `(f(args...))` would be salutary).

veselov:
Thank you.

PR #126, opened 2017-06-16
Don't cast character constants to char*

veselov:
g++ doesn't quite like when character constants are cast to char*


Quuxplusone:
Committed as fc1c658f7fed75ae95f64c641ad6d6192ff1f267 — thanks!

PR #127, opened 2017-06-22
Make HASH_DEL_IN_BKT look more like HASH_ADD_TO_BKT; refactor HASH_DELETE

Quuxplusone:
HASH_DELETE now calls HASH_DELETE_HH, and avoids an ELMT_TO_HH conversion.
Also, seeing that `(head)` in these cases (which really points to a bucket
head, not a table head, so arguably it's confusingly named) can be a really
complicated expression, let's pull it out into a local variable to help
the optimizer.

No intended functional change.

(ping @veselov)


veselov:
Absolutely

PR #130, opened 2017-08-28
Enable non-fatal malloc failures (v2)

Quuxplusone:
If memory allocation fails (i.e., the `uthash_malloc` function returns `NULL`),
the default behavior is to terminate the process by calling `exit(-1)`. This
can be modified by re-defining the `uthash_fatal` macro.

    #undef uthash_fatal
    #define uthash_fatal(msg) my_fatal_function(msg)

The fatal function should terminate the process or `longjmp` back to a safe
place. Note that an allocation failure may leave allocated memory that cannot
be recovered. After `uthash_fatal`, the hash table object should be considered
unusable; it might not be safe even to run `HASH_CLEAR` on the hash table
when it is in this state.

To enable "returning a failure" if memory cannot be allocated, define the
macro `HASH_OOM_OK` before including the `uthash.h` header file. In this
case, `uthash_fatal` is not used; instead, each allocation failure results in
a single call to `uthash_oom(elt)` where `elt` is the address of the element
whose insertion triggered the failure. The default behavior of `uthash_oom`
is a no-op.

    #undef uthash_oom
    #define uthash_oom(elt) perhaps_recover((element_t *) elt)

Before the call to `uthash_oom`, the hash table is rolled back to the state
it was in prior to the problematic insertion; no memory is leaked.
It is safe to `throw` or `longjmp` out of the `uthash_oom` handler.
The `elt` argument will be of the correct pointer-to-element type, unless
`uthash_oom` is invoked from `HASH_SELECT`, in which case it will be of
`void*` type and must be cast before using. In any case, `elt->hh.tbl`
will be `NULL`.

Allocation failure is possible only when adding elements to the hash table
(including the `ADD`, `REPLACE`, and `SELECT` operations).
`uthash_free` never fails.


veselov:
LGTM. I could suggest renaming `HASH_OOM_OK` to `HASH_OOM_RECOVER` (along with the IF_ macro), but I hate longer names as well.

Quuxplusone:
I thought again this morning about switching the behavior based on the presence of `-Duthash_oom(elt)=(my handler)`, but then I realized the real killer of that idea. Right now there are macros `uthash_oom` and `HASH_OOM_OK`; the all-caps one needs to be defined before the first inclusion of uthash.h (just like `HASH_BLOOM` and `HASH_USING_NO_STRICT_ALIASING`), and the lowercase one can be undef'ed and redefined as much as you like (just like `uthash_fatal`, `uthash_bzero`, and so on). If I made the OOM-OK behavior conditional on the definition of the `uthash_oom` macro, then the user would *have* to define `uthash_oom` before the first inclusion of uthash.h — which would be an inconsistency with every other lowercase macro in the library. So okay, we do really need both macros (one all-caps and one lowercase).

I still wonder if I ought to `#warning` or something if `defined(uthash_oom) && !defined(HASH_OOM_OK)`, or if `defined(uthash_fatal) && defined(HASH_OOM_OK)`, but that also might be annoying to the user, so I think I'll wait and see if anyone complains.

I'm inclined to rename `uthash_oom` to `uthash_oom_fyi`, for consistency with `uthash_expand_fyi` and `uthash_noexpand_fyi`. Alternatively: `uthash_nonfatal_oom` for contrast with `uthash_fatal`. Any thoughts on those names?

veselov:
Another idea for renaming `HASH_OOM_OK` to `HASH_OOM_CHECK`?

For the warnings - no strong opinion; I believe that most of the time users will use the default handler, and check OOM condition by testing `hh.tbl`, but it would be interesting to see what really happens.

The proposed lowercase macro names LGTM as well.

Quuxplusone:
Merged 01e5a878d7660e07177a7cb056c769297374d9e3.

PR #137, opened 2018-01-11
resolve gcc 'snprintf’ output truncation warnings

andred:


Quuxplusone:
I've taken 96079f6 immediately — thanks! The other one is likely a true positive as well, but I think the appropriate fix is more likely to be changing `%f` to `%g` rather than increasing the size of the buffer. I'm going to think about it and probably push the patch later this week.

Quuxplusone:
Merged the other patch (the hashscan one) via 3a0c628ff312a5b61456d7d9b5dc96cce5fc943d — thanks for bringing it to my attention!

PR #139, opened 2018-02-06
Use local variable for key length

### LINKED ISSUE #131: Number of calls to uthash_strlen
The convenience macros HASH_FIND_STR, HASH_ADD_STR and HASH_REPLACE_STR call uthash_strlen more frequently than necessary. I would recommend to cache this key length and pass it to HASH_FIND, HASH_ADD and HASH_REPLACE, respectively.

tbeu:
Close #131.


Quuxplusone:
Merged as 29dbbd35851b21ffad93230cb9bd0479e334aec9 — thanks @tbeu!

PR #143, opened 2018-02-15
Resolve MSVC `/Wall /WX` warnings.

silvioprog:
MSVC 15 and later emit the following warnings:

```batch
[snip]
2>c:\utstring.h(209): error C2220: warning treated as error - no 'object' file generated
2>c:\utstring.h(209): warning C4267: '=': conversion from 'size_t' to 'long', possible loss of data
2>c:\utstring.h(286): warning C4267: '=': conversion from 'size_t' to 'long', possible loss of data
2>c:\utstring.h(287): warning C4267: '=': conversion from 'size_t' to 'long', possible loss of data
2>c:\utstring.h(322): warning C4267: '=': conversion from 'size_t' to 'long', possible loss of data
2>c:\utstring.h(328): warning C4267: '=': conversion from 'size_t' to 'long', possible loss of data
2>c:\utstring.h(368): warning C4267: '=': conversion from 'size_t' to 'long', possible loss of data
[snip]
```

This commit fixes them!

### DIFF
--- a/src/utstring.h
+++ b/src/utstring.h
@@ -206,7 +206,7 @@ UTSTRING_UNUSED static void _utstring_BuildTableR(
 {
     long i, j;
 
-    i = P_NeedleLen - 1;
+    i = (long) P_NeedleLen - 1;
     j = i + 1;
     P_KMP_Table[i + 1] = j;
     while (i >= 0)
@@ -283,8 +283,8 @@ UTSTRING_UNUSED static long _utstring_findR(
     long V_FindPosition = -1;
 
     /* Search from right to left. */
-    j = (P_HaystackLen - 1);
-    i = (P_NeedleLen - 1);
+    j = (long) (P_HaystackLen - 1);
+    i = (long) (P_NeedleLen - 1);
     while ( (j >= 0) && (j >= i) )
     {
         while ( (i < (int)P_NeedleLen) && (P_Needle[i] != P_Haystack[j]) )
@@ -319,13 +319,13 @@ UTSTRING_UNUSED static long utstring_find(
 
     if (P_StartPosition < 0)
     {
-        V_StartPosition = s->i + P_StartPosition;
+        V_StartPosition = (long) (s->i + P_StartPosition);
     }
     else
     {
         V_StartPosition = P_StartPosition;
     }
-    V_HaystackLen = s->i - V_StartPosition;
+    V_HaystackLen = (long) (s->i - V_StartPosition);
     if ( (V_HaystackLen >= (long) P_NeedleLen) && (P_NeedleLen > 0) )
     {
         V_KMP_Table = (long *)malloc(sizeof(long) * (P_NeedleLen + 1));
@@ -365,7 +365,7 @@ UTSTRING_UNUSED static long utstring_findR(
 
     if (P_StartPosition < 0)
     {
-        V_StartPosition = s->i + P_StartPosition;
+        V_StartPosition = (long) (s->i + P_StartPosition);
     }
     else
     {


silvioprog:
I forgot some info. It has been happen on MSVC building for Win64.

Quuxplusone:
Hi @silvioprog,

The original sin here seems to have been 38b072a6. Then came 15fd7031, and 2eeb5f42 reverting it. Nobody ever thought to actually make the types *match*, for some reason. But I'm confident that explicit casts are not the most elegant way to solve this problem. Could you take a look at everywhere that currently uses `long` and `size_t` in this file, and decide which of the two types is actually appropriate, and revise this PR to use the appropriate type in each case? (Pro-tip: My rule of thumb is that `long` is *never* appropriate for C code, because its size varies randomly from platform to platform.)

Btw, to review the history of a particular line, you can use

    git blame utstring.h
    git show e4bea624
    git blame e4bea624~ -- utstring.h
    git show 2eeb5f42
    git blame 2eeb5f42~ -- utstring.h

and so on.

silvioprog:
Hello @Quuxplusone , thanks for explaining. I totally agree using `size_t` instead of `long` to represent string length. I'm going to change all `long` occurrences to `size_t` and back with some feedback ...

ps. I'm not sure how to run `utstring` testings. :confused:

Quuxplusone:
You probably figured this out, but fyi to run tests:

    cd tests
    make
    make thorough  # this is optional in that it might not work on your platform

silvioprog:
Finally some progress and a code that works on gcc/clang-Linux(64), gcc-ARM(32), MinGW(32/64) and MSVC(32/64). :sweat_smile: 

The `size_t` is awesome type to represent a buffer size equals, greater than zero or something like that, so `s->i`/`s->n` was kept as is, but a string or entity position is usually represented by a plain `int`, take a look at `stdio` functions:

```bash
...
SYNOPSIS
       #include <stdio.h>

       int printf(const char *format, ...);
       int fprintf(FILE *stream, const char *format, ...);
       int dprintf(int fd, const char *format, ...);
       int sprintf(char *str, const char *format, ...);
       int snprintf(char *str, size_t size, const char *format, ...);

       #include <stdarg.h>

       int vprintf(const char *format, va_list ap);
       int vfprintf(FILE *stream, const char *format, va_list ap);
       int vdprintf(int fd, const char *format, va_list ap);
       int vsprintf(char *str, const char *format, va_list ap);
       int vsnprintf(char *str, size_t size, const char *format, va_list ap);
...
RETURN VALUE
       Upon successful return, these functions return the number of characters
       printed (excluding the null byte used to end output to strings).
```

their returns usually represents a string length por position. so it seems this two commits is the most portable way to fix the reported problem: 05e556c3b94c5884360c3d0c2322630e3a14195f / d9c298e4690529b3c88d8288148ed3cfba45cefe (this last was fixed for MinGW).

Anyway, [I fixed it using `size_t/ssize_t` too](http://brook.no-ip.org/download/0001-utstring.h-change-all-occurrences-from-long-to-size_.patch)(some tests will fail, unfortunately), but since the parts you used `size_t` (except the buffer size, sure) just represents an entity position, the portable type should be `int` instead of `size_t`.

Could you take a look at 05e556c3b94c5884360c3d0c2322630e3a14195f / d9c298e4690529b3c88d8288148ed3cfba45cefe? If you approve them I will send the final PR. :-)

silvioprog:
I have a question, is there any situation that `P_StartPosition`/`P_NeedleLen` reaches the `utstring_len()` size? If so, my patch is wrong. :confused:

Quuxplusone:
Please create a "feature branch" in your clone of this repository:

    git checkout master  # this is where your changes are, at the moment
    git checkout -b msvc-changes  # give it a name that's not "master"

Push the branch to your GitHub:

    git push origin msvc-changes

Then, use GitHub's web interface to open a new pull request *from that branch*. (Do not upload any diff by hand.) This will allow you to update the pull request automatically, simply by pushing improvements to your feature branch.

Once there's a valid pull request available, I can leave review comments on it (and you can address the comments, by pushing improvements to your feature branch, which will automatically update the pull request).

silvioprog:
Thanks a lot for the explanations! I'll try it to send a new PR ...

PR #146, opened 2018-02-16
Small fixes

silvioprog:


src/utstring.h
```diff
@@ -43,6 +43,10 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define oom() exit(-1)
 #endif
 
+#ifndef UTSTRING_INITIAL_SIZE
+#define UTSTRING_INITIAL_SIZE 0x64
+#endif
```
Quuxplusone: (A) If somebody has a strong use-case for tweaking this specific parameter (`UTSTRING_INITIAL_SIZE`), then I might be swayed; but a priori, I'd rather see this particular diff come as part of a larger, comprehensive thinking-out of utstring's allocation machinery. Like, maybe it should be using a primitive such as `utstring_realloc` that the user is allowed to `-D`efine, and maybe we need to think about what realloc-of-size-zero means, and so on. 

(B) Writing `0x64` when you mean `100` is just as confusing as writing `0100` when you mean `64`. I don't know why you'd do that.

src/utstring.h
```diff
@@ -58,33 +62,33 @@ do {                                                       \
...
   free(s);                                                 \
-} while(0)
+} while (0)
 
 #define utstring_new(s)                                    \
 do {                                                       \
-   s = (UT_string*)malloc(sizeof(UT_string));              \
-   if (!s) oom();                                          \
+   (s) = (UT_string*)malloc(sizeof(UT_string));            \
+   if (!(s)) oom();                                        \
```
Quuxplusone: Merged as 29fc26a8 — thanks!

src/utstring.h
silvioprog: (A) 1. On embedded systems (e.g: an AVR/PIC with just 512 of RAM), the programmer could choose a value smaller than 100; On common fast systems, it can be increased (1 kB, 8 kB) via building system.

(B) Oops, sorry. It was just a habit from microcontroller coding!  :sweat_smile:

src/utstring.h
silvioprog: You're welcome dude! :+1:

src/utstring.h
silvioprog: (B) additional info: It uses the same concept: https://github.com/troydhanson/libut/blob/master/src/utvector.c#L29 . (however `INITIAL_SIZE` should be `UTVECTOR_INITIAL_SIZE`)

PR #158, opened 2018-07-15
Update uthash.h

NulAsh:
Removed unnecessary string
Actually, line 228 is unreachable anyway.
HASH_RECORD_OOM(oomed) is just exit(-1)


Quuxplusone:
`git grep HASH_RECORD_OOM` shows another definition for it.

NulAsh:
You mean HASH_NONFATAL_OOM? Then why not exchange lines 227 and 228?

Quuxplusone:
> Then why not exchange lines 227 and 228?

Because in the case that `HASH_RECORD_OOM` *is* just going to exit, there's no point wasting time in a call to the user-defined `uthash_free`.

Issue #157, opened 2018-07-05
Rename uthash_memcmp to HASH_KEYCOMPARE or similar


Quuxplusone:
uthash uses `memcmp` to compare keys. For the benefit of platforms where the native libc contains `bcmp`-but-not-`memcmp`, we provide a customization point `uthash_memcmp` so that the user can compile with

    gcc main.c -Duthash_memcmp=bcmp

However, this is misleading because it implies more constraints than there actually are:
- `bcmp` doesn't actually give the same postconditions as `memcmp`. `bcmp` tests equality and returns only "zero or non-zero", whereas `memcmp` tests ordering and returns "negative, zero, or positive". So `bcmp` is not a drop-in replacement for `memcmp`; yet it *is* still a valid replacement for `uthash_memcmp`.
- [On the mailing list, someone asked](https://groups.google.com/d/msg/uthash/DkQuA2TMxIY/hjCg0QAWCgAJ) how to use uthash with arbitrary struct keys (containing padding, nested pointers, etc). I suggested to `-DHASH_FUNCTION=mystruct_get_hash`. This swaps out the comparison function easily enough; but it won't work out of the box unless you also `-Duthash_memcmp=mystruct_isnt_equal`. Again, `mystruct_isnt_equal` is *even more clearly* not a drop-in replacement for `memcmp` in general; but doing the replacement is the right thing to do in this case (arguably).

So I claim that `uthash_memcmp` is misnamed, and it should have a name that more accurately reflects what it does — for example, `HASH_KEYCOMPARE(k1, k2, keylen)`.

This would also serve the (arguably bad-idea) purposes of users who want to `-DHASH_KEYCOMPARE(...)=0` to "save" the cost of a memcmp on each bucket element (in exchange for occasional wrong results in the presence of hash collisions — collisions that start to become [expected](https://en.wikipedia.org/wiki/Birthday_problem) in the tens-of-thousands-of-elements range).

The question is, if we introduce `HASH_KEYCOMPARE`, what do we do with `uthash_memcmp`? My suggestion is

    #if defined(HASH_KEYCOMPARE)
    /* cool */
    #elif defined(uthash_memcmp)
    #define HASH_KEYCOMPARE(a,b,n) uthash_memcmp(a,b,n)
    #warning "uthash_memcmp is deprecated; please use HASH_KEYCOMPARE instead"
    #else
    #include <string.h>
    #define HASH_KEYCOMPARE(a,b,n) memcmp(a,b,n)
    #endif

[...]

PR #168, opened 2018-12-15
Rename uthash_memcmp to HASH_KEYCOMPARE, step 1.

Quuxplusone:
Addresses the first half of #157.

We merge this, make a minor release (2.1.0),
and then fix up all the tests (e.g. `test6.c`) to redefine
`HASH_KEYCOMPARE` instead of `uthash_memcmp`.

After a few months like that, we eliminate the mention of
`uthash_memcmp` from `uthash.h` and make a major release (3.0.0).

### DIFF
--- a/doc/userguide.txt
+++ b/doc/userguide.txt
@@ -991,6 +991,10 @@ didn't zero-fill the structure, this padding would contain random values.  The
 random values would lead to `HASH_FIND` failures; as two "identical" keys will
 appear to mismatch if there are any differences within their padding.
 
+Alternatively, you can customize the global <<hash_keycompare,key comparison function>>
+and <<hash_functions,key hashing function>> to ignore the padding in your key.
+See <<hash_keycompare,Specifying an alternate key comparison function>>.
+
 [[multilevel]]
 Multi-level hash tables
 ~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


Quuxplusone:
@troydhanson, @veselov: thoughts?

veselov:
I'm fine with this change, that is, it'd be easy to absorb.
However, if there is a freedom to go crazy with comparison function, is there a possibility that function can be somehow specified at time of hash creation? Otherwise, every hash that is different has to be used from an isolated source file...

Quuxplusone:
> So just to make sure I understand. It seems like the user would have to redefine the hash function if they redefine the key comparison function. The requirement would be that, for any keys that compare as equal, the hash function must return the same bucket number for both of them. Otherwise we get structures going into different buckets (due to hash function) though their keys are identical (according to key comparison function).

In general, yes, that's true. Except for the third documented use-case:

> Another reason to substitute your own key comparison function is to trade off correctness for raw speed. During its linear search of a bucket, uthash always compares the 32-bit `hashv` first, and calls `HASH_KEYCOMPARE` only if the `hashv` compares equal. This means that `HASH_KEYCOMPARE` is called at least once per successful find. Given a good hash function, we expect the `hashv` comparison to produce a "false positive" equality only once in four billion times. Therefore, we expect `HASH_KEYCOMPARE` to produce `0` most of the time. If we expect many successful finds, and our application doesn't mind the occasional false positive, we might substitute a no-op comparison function...

That is, if the user can somehow ensure that "`HASH_FUNCTION(a) == HASH_FUNCTION(b)` iff `(a == b)`", then the user can safely define `-DHASH_KEYCOMPARE(a,b,n)=0`.

"uthash always compares the 32-bit `hashv` first..." has been true since 04be2050 (July 2016).

Hmm. I just noticed that prior to 614b6a5, there was a macro named `HASH_KEYCMP`, which makes me wonder if that's the name I should use instead of `HASH_KEYCOMPARE`. On the one hand, clearly nobody is depending on the spelling of a macro that's been gone for 2+ years; on the other hand, maybe `HASH_KEYCMP` _would_ more concisely convey the idea that it should behave like memcmp/bcmp. At the moment I'm actually leaning toward `HASH_KEYCMP`.

PR #169, opened 2018-12-17
Bump version to 2.1.0.

Quuxplusone:
This will be merged after #168.
@tstruk, this is relevant to your interests. :)


Quuxplusone:
For posterity, the recipe for making a release tag is:
```
git tag v2.1.0
git push upstream v2.1.0
```
(This creates a "lightweight" tag, as opposed to an "annotated" tag; all existing tags except for v1.9.8 are "lightweight" and I think that's fine.) No special action is needed to create the listing on https://github.com/troydhanson/uthash/releases ; that happens automatically when a tag with the `vX.Y.Z` format is pushed to GitHub.
https://git-scm.com/book/en/v2/Git-Basics-Tagging

PR #171, opened 2019-03-14
Remove hashing in HASH_FIND when head is null

huansong:
The hashing part in HASH_FIND is not needed when the table does not exist. This mainly helps reduce the costly calculation of HASH_VALUE for certain cases. We found and fixed this issue for [https://github.com/pmodels/mpich](https://github.com/pmodels/mpich).

### DIFF
--- a/src/uthash.h
+++ b/src/uthash.h
@@ -176,8 +176,11 @@ do {                                                                            
 #define HASH_FIND(hh,head,keyptr,keylen,out)                                     \
 do {                                                                             \
   unsigned _hf_hashv;                                                            \
-  HASH_VALUE(keyptr, keylen, _hf_hashv);                                         \
-  HASH_FIND_BYHASHVALUE(hh, head, keyptr, keylen, _hf_hashv, out);               \
+  (out) = NULL;                                                                  \
+  if (head) {                                                                    \
+    HASH_VALUE(keyptr, keylen, _hf_hashv);                                       \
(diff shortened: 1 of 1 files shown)


src/uthash.h
```diff
@@ -176,8 +176,11 @@ do {
 #define HASH_FIND(hh,head,keyptr,keylen,out)                                     \
 do {                                                                             \
   unsigned _hf_hashv;                                                            \
-  HASH_VALUE(keyptr, keylen, _hf_hashv);                                         \
-  HASH_FIND_BYHASHVALUE(hh, head, keyptr, keylen, _hf_hashv, out);               \
+  (out) = NULL;                                                                  \
+  if (head) {                                                                    \
+    HASH_VALUE(keyptr, keylen, _hf_hashv);                                       \
+    HASH_FIND_BYHASHVALUE(hh, head, keyptr, keylen, _hf_hashv, out);             \
+  }                                                                              \
```
Quuxplusone: LGTM, but I think I prefer to see `(out) = NULL` executed in an `else` branch, just in case `out` has side effects.

huansong: Thanks. I agree. Then we also need to move `(out) = NULL` to an `else` branch in `HASH_FIND_BYHASHVALUE`, right?

Quuxplusone: I guess `out` is actually used as a loop control variable in `HASH_FIND_IN_BKT`, so it will get assigned many times regardless of what we do up here. Maybe this is better, then, because it's less work for the compiler. I'll merge this.

PR #172, opened 2019-04-15
Fix issues reported by PVS-Studio

silvioprog:
* V522 There might be dereferencing of a potential null pointer 'str->buf'.
* V522 There might be dereferencing of a potential null pointer '(str->buf)->d'.

### DIFF
--- a/src/utstring.h
+++ b/src/utstring.h
@@ -64,7 +64,7 @@ do {                                                       \
 do {                                                       \
   (s)->n = 0; (s)->i = 0; (s)->d = NULL;                   \
   utstring_reserve(s,100);                                 \
-  (s)->d[0] = '\0';                                        \
+  if ((s)->d) (s)->d[0] = '\0';                            \
 } while(0)
 
 #define utstring_done(s)                                   \
@@ -82,8 +82,8 @@ do {                                                       \
 #define utstring_new(s)                                    \
 do {                                                       \
    (s) = (UT_string*)malloc(sizeof(UT_string));            \
-   if (!(s)) oom();                                        \
-   utstring_init(s);                                       \
+   if ((s)) utstring_init(s);                              \
+   else oom();                                             \
 } while(0)
 
 #define utstring_renew(s)                                  \
--- a/src/utstring.h
+++ b/src/utstring.h
@@ -82,7 +82,7 @@ do {                                                       \
 #define utstring_new(s)                                    \
 do {                                                       \
    (s) = (UT_string*)malloc(sizeof(UT_string));            \
-   if ((s)) utstring_init(s);                              \
+   if (s) utstring_init(s);                                \
    else oom();                                             \
 } while(0)
 


src/utstring.h
```diff
@@ -82,8 +82,8 @@ do {                                                       \
 #define utstring_new(s)                                    \
 do {                                                       \
    (s) = (UT_string*)malloc(sizeof(UT_string));            \
-   if (!(s)) oom();                                        \
-   utstring_init(s);                                       \
+   if ((s)) utstring_init(s);                              \
+   else oom();                                             \
```
Quuxplusone: Why doesn't PVS-Studio also complain about the similar use of `oom` in `utstring_reserve`/`bincpy`/`concat`? Is it just that your particular usage pattern doesn't exercise that codepath in `utstring_reserve`?

I would expect PVS-Studio to know that `oom()` means `exit(-1)` which is a noreturn function.

src/utstring.h
```diff
@@ -82,8 +82,8 @@ do {                                                       \
 #define utstring_new(s)                                    \
 do {                                                       \
    (s) = (UT_string*)malloc(sizeof(UT_string));            \
-   if (!(s)) oom();                                        \
-   utstring_init(s);                                       \
+   if ((s)) utstring_init(s);                              \
```
Quuxplusone: Redundant parens in `if ((s))` — should be just `if (s)`.

src/utstring.h
silvioprog: Oh, I forgot to mention an important information: it happens only if we disable the `oom`, e.g.:

```c
#ifndef oom
#define oom() (void) 0
#endif
```

src/utstring.h
silvioprog: Hm... I've followed the same used in https://github.com/troydhanson/uthash/blob/master/src/utstring.h#L54, https://github.com/troydhanson/uthash/blob/master/src/utstring.h#L85.

src/utstring.h
Quuxplusone: Ah. Well, if you disable `oom()`, then of course you have a bug, and PVS-Studio is right to warn you about it. The solution is to not do that.

src/utstring.h
Quuxplusone: FYI, the difference there is that those single sets of parens are actually doing something. There's no instance of `((s))` in what you linked to — just `(s)`.

src/utstring.h
silvioprog: So, I wouldn't like to exit my application if oom... in this case, what do you recommend? :confused:

src/utstring.h
silvioprog: Interesting information. I'm going to fix it to upgrade the PR ...

silvioprog: Removed the redundant parenthesis in `if`.

src/utstring.h
silvioprog: It would be nice if the `utstring` could follow the same idea as `HASH_NONFATAL_OOM` from `uthash` to avoid to exit the application if OOM ...

silvioprog:
Any chance to apply the changes? :confused:

Quuxplusone:
It seems like the only two diffs left in this PR are handling "`s->d` might be null, if `oom()` is a no-op"; but `oom()` shouldn't ever _be_ a no-op — it should `exit` or `abort` or `longjmp` — and so I think there's nothing left to do here. I'm closing this; if there are any remaining/new static-analyzer issues, please open a new issue/PR.

PR #175, opened 2019-05-06
Rename oom() to utarray_oom() and document how to override it.

### LINKED ISSUE #174: oom() calls exit(): This makes this library unfit for usage in shared libraries
For example,

https://stackoverflow.com/questions/14199689/how-can-i-handle-exit-calls-in-3rd-party-library-code

editorconfig/editorconfig-core-c#12

xuhdev:
Fix #174

### DIFF
--- a/doc/utarray.txt
+++ b/doc/utarray.txt
@@ -370,4 +370,12 @@ Notes
    For this reason, it's usually better to refer to an element by its integer
    'index' in code whose duration may include element insertion.
 
+6. To override the default out-of-memory handling behavior (which calls `exit(-1)`),
+   override the `utarray_oom` macro before including `utarray.h` or functions that use it.
+   For example,
+
+  #define utarray_oom() do { set_error_code(100); } while (0)
+  ...
+  #include "utarray.h"
+
 // vim: set nowrap syntax=asciidoc:
--- a/src/utarray.h
+++ b/src/utarray.h
@@ -38,8 +38,14 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define UTARRAY_UNUSED
 #endif
 
-#ifndef oom
-#define oom() exit(-1)
+/* Backward compatibility of oom */
+#ifdef oom
+#error "The name of macro 'oom' is deprecated and has been changed to 'utarray_oom'. Please update your code."
+#define utarray_oom() oom()
+#endif
+
+#ifndef utarray_oom
+#define utarray_oom() exit(-1)
 #endif
 
 typedef void (ctor_f)(void *dst, const void *src);
@@ -78,7 +84,9 @@ typedef struct {
 
 #define utarray_new(a,_icd) do {                                              \
   (a) = (UT_array*)malloc(sizeof(UT_array));                                  \
-  if ((a) == NULL) oom();                                                     \
+  if ((a) == NULL) {                                                          \
+    utarray_oom();                                                            \
+  }                                                                           \
   utarray_init(a,_icd);                                                       \
 } while(0)
 
@@ -92,7 +100,9 @@ typedef struct {
     char *utarray_tmp;                                                        \
     while (((a)->i+(by)) > (a)->n) { (a)->n = ((a)->n ? (2*(a)->n) : 8); }    \
     utarray_tmp=(char*)realloc((a)->d, (a)->n*(a)->icd.sz);                   \
-    if (utarray_tmp == NULL) oom();                                           \
+    if (utarray_tmp == NULL) {                                                \
+      utarray_oom();                                                          \
+    }                                                                         \
     (a)->d=utarray_tmp;                                                       \
   }                                                                           \
 } while(0)


doc/utarray.txt
```diff
@@ -370,4 +370,11 @@ Notes
    For this reason, it's usually better to refer to an element by its integer
    'index' in code whose duration may include element insertion.
 
+6. To override the default out-of-memory handling behavior (which calls `exit(-1)`),
+   override the `utarray_oom` macro before including `utarray.h`. For example,
+
+  #define utarray_oom() set_error_code(100);
```
Quuxplusone: You've got an extra semicolon here. And really we should probably encourage macro hygiene:

    #define utarray_oom() do { set_error_code(100); } while (0)

src/utarray.h
```diff
@@ -38,8 +38,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define UTARRAY_UNUSED
 #endif
 
-#ifndef oom
-#define oom() exit(-1)
+#ifndef utarray_oom
+#define utarray_oom() exit(-1)
 #endif
```
Quuxplusone: Is anyone out there using `-Doom()=whatever` today? If so, we might need to add something like

    #ifdef oom
    #error "The name of macro 'oom' has been changed to 'utarray_oom'. Please update your code."
    #endif

Does anyone reading this PR want to speak up and say "I've used `-Doom()=`?"
I'm inclined to leave this PR open for comments for a month, and then if nobody complains, merge it without any `#ifdef oom` safety net. `oom` was a terrible name for a macro.

src/utarray.h
```diff
@@ -92,7 +92,7 @@ typedef struct {
     char *utarray_tmp;                                                        \
     while (((a)->i+(by)) > (a)->n) { (a)->n = ((a)->n ? (2*(a)->n) : 8); }    \
     utarray_tmp=(char*)realloc((a)->d, (a)->n*(a)->icd.sz);                   \
-    if (utarray_tmp == NULL) oom();                                           \
+    if (utarray_tmp == NULL) utarray_oom();                                   \
```
Quuxplusone: Here and above, I would like to see curly braces around the `if` body, as long as we're updating the code.

    if (utarray_tmp == NULL) { \
        utarray_oom();         \
    }                          \

src/utarray.h
xuhdev: How about add them for now and remove them after a month if no one complains?

xuhdev:
I've made an update that has fixed most issues, except the one I commented on.

src/utarray.h
Quuxplusone: Well, if I add the `#ifdef oom` safety net, I'm going to leave it in at least for one release.
I have no real sense of how many people use uthash straight from github master, versus from distributions that update `/usr/include/uthash/uthash.h`, versus vendoring it into their codebase and updating it once every couple of years, if ever.

But I guess adding the safety net is the right thing to do.

xuhdev: OK! Please see my updated commit.

xuhdev:
If there is no problem with this issue, can you merge it? The changes shouldn't break any compatibility with existing APIs, in particular `oom`.

doc/utarray.txt
```diff
@@ -370,4 +370,12 @@ Notes
    For this reason, it's usually better to refer to an element by its integer
    'index' in code whose duration may include element insertion.
 
+6. To override the default out-of-memory handling behavior (which calls `exit(-1)`),
+   override the `utarray_oom` macro before including `utarray.h` or functions that use it.
+   For example,
+
+  #define utarray_oom() do { set_error_code(100); } while (0)
```
Quuxplusone: Notice that if `utarray_oom()` is defined to something that does NOT either terminate execution or `longjmp` away from the trouble zone, then `utarray_new` will dereference a null pointer and segfault.

I will adjust this documentation to reflect that.

But, knowing that, does `utarray_oom` still solve your underlying problem? Or is it the wrong tool for the job?

In particular, should you instead be looking for a way to redefine the meaning of `malloc` in this code, so that it can't fail in the first place? uthash also provides `uthash_malloc` and `uthash_free`.

src/utarray.h
```diff
@@ -78,7 +84,9 @@ typedef struct {
 
 #define utarray_new(a,_icd) do {                                              \
   (a) = (UT_array*)malloc(sizeof(UT_array));                                  \
-  if ((a) == NULL) oom();                                                     \
+  if ((a) == NULL) {                                                          \
+    utarray_oom();                                                            \
```
xuhdev: How about adding a `break;` after this line? Then we don't need to adjust the documentation and I can't imagine anything we would lose.

doc/utarray.txt
xuhdev: Yes, it solves my problem. Thanks for asking.

src/utarray.h
Quuxplusone: Adding an `else` would be much clearer than adding a `break`; but I hesitate to do either one, because I'm really not trying to encourage people to write non-fatal OOM handlers. Supporting "non-fatal OOM" in uthash.h was a massive undertaking for (AFAIK) very little benefit. I'd rather people just avoid using `malloc`-that-can-fail in C to begin with.

xuhdev: OK. But what would be your suggestion for shared libraries? They are not supposed to be fatal for the whole program at any time. I think your point is fair, and the default already supports your view. It's just that shared libraries have no choice and utarray offers a solution for them.

Quuxplusone: I think the easiest thing to do is "ignore the problem; if you're using a shared library you are probably on Linux (where [malloc never fails](https://stackoverflow.com/questions/16674370/why-does-malloc-or-new-never-return-null)) or some other hosted desktop system (where malloc never fails AFAIK)."

The most correct thing to do would be to substitute in your own `malloc` which allocates out of a fixed-size memory resource created at the start of the subroutine that wants to use uthash/utarray. Then, on malloc failure (i.e. inside `utarray_oom()`), you `longjmp` back to the beginning of the subroutine, ["wink out"](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0089r0.pdf) your memory resource (so that it doesn't matter whether utarray leaks memory on failure), and continue via your fallback path.

xuhdev: I understand your concerns now and I agree with it. To achieve the goal, instead of plainly leaving out `break;`, how about adding one line that omits a warning message, so as to remind users that they should either write a fatal OOM or do whatever they want to skip the initialization line? I'll also update the document for this.

xuhdev: Just a friendly reminder: Your link also says that, on Linux, `overcommit_memory` can be turned off to make `malloc` failable again. Therefore, I can't agree with your view with respect to "`malloc` never fails"

Quuxplusone:
Merged as cfc95dbe14b8aa47cee0292158876dd118903e48 — thanks!

PR #180, opened 2019-07-29
Fix clang static analyzer false positive

### LINKED ISSUE #128: Clang reports use after free when deleting all hashmap contents
The clang reports "use after free" when analyzing the following piece of code:

```
#include <stdio.h>   /* gets */
#include <stdlib.h>  /* atoi, malloc */
#include <string.h>  /* strcpy */
#include "uthash.h"

struct my_struct {
    int id;                    /* key */
    char name[10];
    UT_hash_handle hh;         /* makes this structure hashable */
};

struct my_struct *users = NULL;

void add_user(int user_id, char *name) {
    struct my_struct *s;

    HASH_FIND_INT(users, &user_id, s);  /* id already in the hash? */
    if (s==NULL) {
      s = (struct my_struct*)malloc(sizeof(struct my_struct));
      s->id = user_id;
      HASH_ADD_INT( users, id, s );  /* id: name of key field */
    }
    strcpy(s->name, name);
}

void delete_all() {
  struct my_struct *current_user, *tmp;

  HASH_ITER(hh, users, current_user, tmp) {
    HASH_DEL(users, current_user);  /* delete it (users advances to next) */
    free(current_user);             /* free it */
  }
}


int main(int argc, char *argv[]) {
    char in[10];
    int id=1, running=1;
    struct my_struct *s;
    unsigned num_users;

    add_user(1, "111111");
    add_user(2, "222222222");
    add_user(3, "3");
    
    delete_all();  /* free any structures */
    return 0;
}
```

Clang report:

```
uthash.c:30:5: warning: Use of memory after it is freed
    HASH_DEL(users, current_user);  /* delete it (users advances to next) */
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./uthash.h:421:5: note: expanded from macro 'HASH_DEL'
    HASH_DELETE(hh,head,delptr)
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~
./uthash.h:369:5: note: expanded from macro 'HASH_DELETE'
    HASH_DELETE_HH(hh, head, &(delptr)->hh)
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./uthash.h:376:17: note: expanded from macro 'HASH_DELETE_HH'
    uthash_free((head)->hh.tbl->buckets,                                         \
    ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./uthash.h:89:34: note: expanded from macro 'uthash_free'
#define uthash_free(ptr,sz) free(ptr)     /* free fcn                        */
                                 ^
uthash.c:30:5: warning: Use of memory after it is freed
    HASH_DEL(users, current_user);  /* delete it (users advances to next) */
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./uthash.h:421:5: note: expanded from macro 'HASH_DEL'
    HASH_DELETE(hh,head,delptr)
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~
./uthash.h:369:5: note: expanded from macro 'HASH_DELETE'
    HASH_DELETE_HH(hh, head, &(delptr)->hh)
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./uthash.h:382:23: note: expanded from macro 'HASH_DELETE_HH'
    if (_hd_hh_del == (head)->hh.tbl->tail) {                                    \
                      ^~~~~~~~~~~~~~
2 warnings generated.
```

inye:
Fixes #128. Clang static analyzer was thinking that it's possible to
free() list head without updating head pointer. This assert assures
static analyzer that branch without head pointer update could be taken
only when we are freeing non-head element.

### DIFF
--- a/src/uthash.h
+++ b/src/uthash.h
@@ -473,6 +473,7 @@ do {                                                                            
       (head)->hh.tbl->tail = HH_FROM_ELMT((head)->hh.tbl, _hd_hh_del->prev);     \
     }                                                                            \
     if (_hd_hh_del->prev != NULL) {                                              \
+      assert(&(head)->hh != _hd_hh_del);                                         \
       HH_FROM_ELMT((head)->hh.tbl, _hd_hh_del->prev)->next = _hd_hh_del->next;   \
     } else {                                                                     \
       DECLTYPE_ASSIGN(head, _hd_hh_del->next);                                   \


Scorbutics:
Hello,
I'm quite interested in this Merge request in order to keep scan-build quiet on the uthash part.
Are there any plan of merging it soon ?
Thank you

src/uthash.h
```diff
@@ -473,6 +473,7 @@ do {
       (head)->hh.tbl->tail = HH_FROM_ELMT((head)->hh.tbl, _hd_hh_del->prev);     \
     }                                                                            \
     if (_hd_hh_del->prev != NULL) {                                              \
+      assert(&(head)->hh != _hd_hh_del);                                         \
```
Quuxplusone, line 476: My concern with this one is that it's introducing `assert` (and `<assert.h>`) into "uthash.h", which historically hasn't been the case. I know "utlist.h" uses `<assert.h>`, but "uthash.h" itself has a lot more users than "utlist.h", so I'm leery of introducing new header dependencies (even if they are C89 standard) and especially new runtime dependencies (e.g. you're introducing a hidden dependency on the libc's `abort` function, right?).

Is there any way to silence scan-build's warning here with core-language features instead of `assert`?

I even wonder if we could just change line 475 from `if (_hd_hh_del->prev != NULL)` to `if (_hd_hh_del != &(head)->hh)` — does that work? It shouldn't really be any less efficient, since we have to load `(head)->hh` on line 476 anyway.

Quuxplusone (changes requested):

src/uthash.h
Quuxplusone: Btw, I'd also be extremely interested in PRs against our `.travis.yml` file so that we could actually _test_ the scan-build build against regressions in this area. Any takers?

chaitu-tk: > I even wonder if we could just change line 475 from if (_hd_hh_del->prev != NULL) to if (_hd_hh_del != &(head)->hh) — does that work?

Nope, this introduced a new warning `warning: Dereference of null pointer [clang-analyzer-core.NullDereference]`.

Using the assert and including <assert.h> in uthash.h solved the problem, can we go ahead with that fix?

Quuxplusone: > including <assert.h> in uthash.h solved the problem, can we go ahead with that fix?

No.

Any volunteers to add scan-build into `.travis.yml`?

chaitu-tk: I can take a stab

Quuxplusone: Now that I'm able to reproduce locally (thanks @chaitu-tk!) I see that we could silence the false positives on test15, test40, and test59 by adding `assert((delptr) != (head))` at the end of `HASH_DELETE` — i.e., "after deleting an item from the hash table, the deleted item is no longer in the table at all, let alone at the _front_ of the table."

Except that this breaks test18, because we want to permit `HASH_DEL(table, table)` to mean "delete the first item in the table." And anyway, I still don't want to add a dependency on `<assert.h>` to uthash.

I've just fixed some low-hanging fruit in 45af88cd1fdc16743fa6d2e763af740e49eda1c5 and f0e1bd9b609f9cf22b779a1ec019ee13d1c85728 (which I'll merge up to this repo once Travis passes).

PR #185, opened 2019-10-29
tests/do_tests.mingw: fix error that .exe's cannot be found

jannick0:
Essentially replace `test.exe` by `./test.exe`, such that `bash` can find it.

### DIFF
--- a/tests/do_tests.mingw
+++ b/tests/do_tests.mingw
@@ -5,7 +5,7 @@ echo "MinGW test script starting"
 for f in test*.exe
 do
   t=`echo $f | sed s/.exe//`
-  `$f > $t.out`
+  "./$f" > "$t.out"
   diff -qb "$t.out" "$t.ans"
   if [ $? -eq 1 ]
   then
--- a/tests/do_tests.mingw
+++ b/tests/do_tests.mingw
@@ -6,14 +6,7 @@ for f in test*.exe
 do
   t=`echo $f | sed s/.exe//`
   "./$f" > "$t.out"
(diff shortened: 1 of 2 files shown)


tests/do_tests.mingw
```diff
@@ -5,15 +5,8 @@ echo "MinGW test script starting"
...
-  diff -qb "$t.out" "$t.ans"
-  if [ $? -eq 1 ]
-  then
-    echo "$f failed"
-  else
-    true # can't have empty else
-    #echo "$f passed"
-  fi
+  "./$f" > "$t.out"
+  diff -b "$t.out" "$t.ans" > "$t.diff" && rm "$t.diff" || echo "$f failed"
```
Quuxplusone: Is this change needed? Does MinGW Bash not support `$?` or something?

(The one-line change of `$f > $t.out` to `"./$f" > "$t.out"` seems fine. I don't know why the old line originally had backticks around it.)

Quuxplusone:
Merged 6c1182dd736cd57d63c4c826adbacb9f2fe85f8f.

I'm still skeptical of the other commit in this PR. It doesn't seem like a burden to me for the user to have to run `diff x.ans x.out`; and it doesn't seem like a *non*-burden for the user to have to `rm *.diff` afterward (and remember not to put any important files named `*.diff` in this directory because now they might get deleted by `make clean`). Also, none of the other platforms' `do_tests` scripts generate intermediate `.diff` files. So I'm inclined to reject that half of this change, and close out this PR.

Issue #186, opened 2019-11-15
Clang warning "increases required alignment" with -Wcast-align


ravone:
When using macro `HASH_ADD_INORDER` I get warning:
```
warning: cast from 'char *' to 'UT_hash_handle *' (aka 'struct UT_hash_handle *') increases required alignment from 1 to 8
uthash.h:405:3: note: expanded from macro 'HASH_ADD_INORDER'
uthash.h:398:3: note: expanded from macro 'HASH_ADD_KEYPTR_INORDER'
uthash.h:377:5: note: expanded from macro 'HASH_ADD_KEYPTR_BYHASHVALUE_INORDER'
uthash.h:303:24: note: expanded from macro 'HASH_AKBI_INNER_LOOP'
uthash.h:147:32: note: expanded from macro 'HH_FROM_ELMT'
```
Perhaps, it's enough just fix this macro
```
#define HH_FROM_ELMT(tbl,elp) ((UT_hash_handle *)(((char*)(elp)) + ((tbl)->hho)))
```
to
```
#define HH_FROM_ELMT(tbl,elp) ((UT_hash_handle *)(void*)(((char*)(elp)) + ((tbl)->hho)))
```

Quuxplusone:
#177 is related.
Are you getting `-Wcast-align` from `-Wall`, or are you deliberately enabling it for some actual reason (and if so, is it warning you _correctly_ about some _real_ issue that should not be ignored on your platform)?

ravone:
@Quuxplusone  I get it from `-Wall`, no any other actual reason. Yeah, it does not seem to be real issue in my code and probably I can ignore it. But maybe we can cast through `(void*)` and make analyzer happy?

Pull request is totally related and I think that you can close this issue.

P.S. Thank you for your work of maintaining this project!

Quuxplusone:
@ravone: I just checked Clang's source code, and I don't see any way to trigger that warning except to pass `-Wcast-align` explicitly (or to pass the magic driver option `-Weverything`, [which you shouldn't expect to be clean](https://quuxplusone.github.io/blog/2018/12/06/dont-use-weverything/)). It's definitely not part of `-Wall` in Clang trunk, anyway.

[...]

PR #187, opened 2019-11-23
Remove all support for MurmurHash (HASH_MUR, HASH_USING_NO_STRICT_ALIASING)

Quuxplusone:
This is related to #186: `HASH_MUR` generates a lot of uthash's `-Wcast-align` noise. There's a lot of crufty technical debt around `HASH_MUR`; I don't think it's pulling its weight, and will continue to think so unless somebody steps up to complain about this pull request.

I think that if someone wants to use MurmurHash with uthash, the right way to do it is to define a function (perhaps an `inline` function) that implements MurmurHash, and then define `HASH_FUNCTION` to the name of that function. I don't think we need to continue supporting MurmurHash out of the box, with all the processor-specific `#ifdef`s and compiler-specific `-fno-strict-aliasing` cruft that this old implementation entails.

Attn: @troydhanson, @archerian, @al-heisner (maybe)


troydhanson:
I’m 100% ok with this. I appreciate your initiative.

Troy

al-heisner:
I think I still have those emails from way back in 2009 where I was messing
around and added murmurhash to uthash (incl. testing on Sparc).
I'm still using uthash in some of my projects and I agree that murmur
doesn't necessarily need to be carried if it's current implementation is
more work than the reward.
I would consider refactoring it with the suggested implementation technique
if I felt really ambitious but I suspect the other available hash
algorithms will serve my needs.

PR #191, opened 2019-12-22
Honor user's CFLAGS

noloader:
The project's required `CFLAGS` are gathered in `UTH_CFLAGS` and used in the target recipes. It leaves `CFLAGS` for the user. Also see https://www.gnu.org/prep/standards/html_node/Command-Variables.html.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -13,22 +13,22 @@ PROGS = test1 test2 test3 test4 test5 test6 test7 test8 test9   \
         test74 test75 test76 test77 test78 test79 test80 test81 \
         test82 test83 test84 test85 test86 test87 test88 test89 \
         test90 test91 test92 test93 test94 test95
-CFLAGS += -I$(HASHDIR)
-#CFLAGS += -DHASH_BLOOM=16
-#CFLAGS += -O2
-CFLAGS += -g
-#CFLAGS += -Wstrict-aliasing=2
-CFLAGS += -Wall
-#CFLAGS += -Wextra
-#CFLAGS += -std=c89
-CFLAGS += ${EXTRA_CFLAGS}
+UTH_CFLAGS += -I$(HASHDIR)
+#UTH_CFLAGS += -DHASH_BLOOM=16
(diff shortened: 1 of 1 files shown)


Quuxplusone:
This seems reasonable; I like the guidelines in https://www.gnu.org/prep/standards/html_node/Command-Variables.html . However:

- this now needs rebasing (thanks for reminding me to merge the no-more-murmurhash patch!)
- if we're putting the user's CFLAGS at the end of the line, then should we put the user's CPPFLAGS (preprocessor flags) at the end of the line as well?
- `s/UTH_CFLAGS/HASH_CFLAGS/` — let's keep the number of prefixes we use to a minimum

tests/Makefile
```diff
@@ -13,22 +13,22 @@ PROGS = test1 test2 test3 test4 test5 test6 test7 test8 test9   \
...
-CFLAGS += ${EXTRA_CFLAGS}
+UTH_CFLAGS += -I$(HASHDIR)
+#UTH_CFLAGS += -DHASH_BLOOM=16
+#UTH_CFLAGS += -O2
+UTH_CFLAGS += -g
+#UTH_CFLAGS += -Wstrict-aliasing=2
+UTH_CFLAGS += -Wall
+#UTH_CFLAGS += -Wextra
+#UTH_CFLAGS += -std=c89
+UTH_CFLAGS += ${EXTRA_CFLAGS}
```
Quuxplusone: Besides my high-level comments in the "Conversation" tab: should this now use `${CFLAGS}` instead of `${EXTRA_CFLAGS}`? I believe the purpose of `EXTRA_CFLAGS` was so that the user can pass their own extra stuff, but if we're going to respect their `CFLAGS` then they don't need this extra mechanism.

Quuxplusone:
It's been >4 years; I'm going to close this PR as inactive. However, I've just put up the very similar #271; if its CI is green, I'm inclined to go ahead with it (leaving a few weeks for other folks to weigh in if they care).

PR #193, opened 2020-01-01
Add the missing HASH_REPLACE_KEYPTR

PantherJohn:


Quuxplusone:
Seems reasonable to me, but needs new tests for each of the newly added macros.

PantherJohn:
@Quuxplusone then run the test and see if you would get any problems.

Issue #198, opened 2020-04-10
VS 2019 gives compiler warning C26451 for UT Array


guywyers:
Similar to issue [195](https://github.com/troydhanson/uthash/issues/195) for UT Hash:

`Warning C26451	Arithmetic overflow: Using operator '*' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '*' to avoid overflow (io.2).`

It seems in this case it's the following line causing the warning:
`#define _utarray_eltptr(a,j) ((void*)((a)->d + ((a)->icd.sz * (j))))`

Doesn't seem to be dramatic, but would be nice to have it fixed.

Quuxplusone:
Can you post a very short C source file that demonstrates the issue?

guywyers:
Here is a snippet:

```
static const UT_icd ut_int64_icd UTARRAY_UNUSED = { sizeof(uint64_t),NULL,NULL,NULL };

int main(int argc, char* argv[])
{
	UT_array* part1;
	UT_array* part2;

	utarray_new(part1, &ut_int64_icd);
	utarray_new(part2, &ut_int64_icd);

	uint64_t i;
	for (i = 1; i < 5; ++i)
		utarray_push_back(part1, &i);

	for (; i < 20; ++i)
		utarray_push_back(part2, &i);

	utarray_concat(part1, part2);
}
```
The last line triggers the warning.
I have spend a bit of time tracing it back and ended up in the `_utarray_eltptr` macro. However, as I said, I do not consider it "life-threatening" so I gave up after about 30 min, so I could be wrong.

Quuxplusone:
Darn, I thought that with a test case I'd be able to reproduce it in Godbolt, but it seems not. https://godbolt.org/z/x-HB64

Can you simplify that test case at all? For example, is this source file enough to trigger the warning?

    #include "utarray.h"
    void test(UT_array* part1) {
        int i;
        utarray_push_back(part1, &i);
    }

guywyers:
Nope, that snippet doesn't trigger the warning. Maybe it's some compiler or preprocessor option which is causing this?

guywyers:
@Quuxplusone: the second snippet triggered the warning.
My comiler options are:
[...]

PR #203, opened 2020-05-10
Fixed issue #198 on master

guywyers:


src/utarray.h
```diff
@@ -58,7 +58,7 @@ typedef struct {
 } UT_icd;
 
 typedef struct {
-    unsigned i,n;/* i: index of next available slot, n: num slots */
+    size_t i,n;/* i: index of next available slot, n: num slots */
```
Quuxplusone, line 61: I'd prefer not to change the layout of `struct UT_array` if we can help it.

I infer that the rest of the changes in this patch are cascading from this one: since you changed the type of `i`, the expression `_ut_i < (a)->i` started warning, so you also had to change the type of `_ut_i`, and so on.

Suppose you leave this line alone, and approach the problem from the other end: where are the individual expressions that trigger the MSVC diagnostic in master, and can we just add casts or other local changes to silence the warnings down there?

guywyers: Actually changing the struct was the **last** thing I did. I tried your suggestion, ie patching individual lines with casts, to no avail.
So in the end I gave up and did a systematic search replace of `unsigned` / `size_t` and that fixed it.
I do agree that changing the struct just doesn't seem right, so I'll have another go at it.

AKZ98 (approved):

AKZ98 (approved):

PR #207, opened 2020-06-25
Support spaces in $exe path

kaloyan13:
If uthash is cloned under a path with spaces, it doesn't run the keyscan.* files.

### DIFF
--- a/tests/keystats
+++ b/tests/keystats
@@ -12,9 +12,11 @@ sub usage {
 
 usage if ((@ARGV == 0) or ($ARGV[0] eq '-h'));
 
-my @exes = glob "$FindBin::Bin/keystat.???";
+my @exes = glob "'$FindBin::Bin/keystat.???'";
+
 my %stats;
 for my $exe (@exes) {
+    $exe =~ s/\ /\\ /g;
     $stats{$exe} = `$exe @ARGV`;
     delete $stats{$exe} if ($? != 0); # omit hash functions that fail to produce stats (nx)
 }


Quuxplusone:
Good catch!

However, I think "manually escaping a path via string replacement" is a security smell. That is, I think the real fix would be to replace

    `$exe @ARGV`

with something in terms of [`system(@args)`](https://perldoc.perl.org/functions/system.html). Do you happen to know Perl well enough to figure that out for me? :)

I've never used the `keystats` program; I'd be mildly interested to hear about your use-case.

kaloyan13:
I can't smell anything :)
system() gives no output. You will have to pipe output, then parse.
The other option is [open("-|")](https://perldoc.perl.org/functions/open.html), but for a single line, its too much code.

I used keystats to see, if there is a faster/better function than the default one.
fcn  ideal%     #items   #buckets  dup%  fl   add_usec  find_usec  del-all usec
---  ------ ---------- ---------- -----  -- ---------- ----------  ------------
SAX   91.4%     323765     262144   69%  ok     369341     301862         37661
JEN   91.2%     323765     262144   69%  ok     331568     290620         33546
OAT   90.9%     323765     262144   69%  ok     504209     462986         32241
FNV   90.3%     323765     262144   69%  ok     377140     330207         31727
BER   90.0%     323765     262144   69%  ok     276762     252049         35497
SFH   87.8%     323765     262144   69%  ok     244188     197380         36147

SAX shows 0.2% ideal, but SFH is the fastest, as I can see.

Quuxplusone:
Merged 7f0aadbb84689d09f46fb7d41b5c9547fb642892 — thanks!

PR #212, opened 2020-12-02
Always include <stdint.h>, unless HASH_NO_STDINT is defined by the user.

Quuxplusone:
See 762026ce and b1d8ab07 for previous attempts (circa 2014-2015) to detect whether it's OK to use <stdint.h>; and see #211 for how that logic is still incomplete.

Rather than continue down this path, how about we just say that either your system should be C99 (it's the year 2020!), _or_ you should provide your own "polyfill" version of `<stdint.h>` in the normal search path, _or_ you should `-DHASH_NO_STDINT` and figure out the details yourself (e.g. by providing the necessary typedefs above where you `#include <uthash.h>`).

Also support `-DHASH_DEFINE_OWN_STDINT` for this next release, so that there's an easy backward-compatibility mode. I might rip out this mode in the release after next, though.

@tbeu: please take a look, since you're the author of much of this logic (back in 2014-2015). Any objection to my thus ripping it out?

### DIFF
--- a/src/uthash.h
+++ b/src/uthash.h
@@ -30,6 +30,14 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <stddef.h>   /* ptrdiff_t */
 #include <stdlib.h>   /* exit */
 
+#if defined(HASH_NO_STDINT) && HASH_NO_STDINT
+#elif defined(HASH_DEFINE_OWN_STDINT) && HASH_DEFINE_OWN_STDINT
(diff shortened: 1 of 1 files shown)


tbeu:
OK, will have to recap...

Quuxplusone:
Merged 15ad04278f19adefad34353143d944cb0c007fe2.

tbeu:
Agree to remove my old contribution from a library header.

FYI, here are the changes that need to be done on the application side when migrating to uthash v2.2.0: https://github.com/modelica/ModelicaStandardLibrary/pull/3688

PR #213, opened 2020-12-03
Fix broken example.c

lnslbrty:
* replaced usage of dangerous gets() with fgets()
 * replaced strcpy with strncpy

Signed-off-by: Toni Uhlig <matzeton@googlemail.com>

### DIFF
--- a/tests/example.c
+++ b/tests/example.c
@@ -1,11 +1,14 @@
+#include <assert.h>
 #include <stdio.h>   /* gets */
 #include <stdlib.h>  /* atoi, malloc */
 #include <string.h>  /* strcpy */
 #include "uthash.h"
 
+#define NAME_MAXLEN 10
+
 struct my_struct {
     int id;                    /* key */
-    char name[10];
+    char name[NAME_MAXLEN];
     UT_hash_handle hh;         /* makes this structure hashable */
 };
 
@@ -21,6 +24,7 @@ void add_user(int user_id, char *name)
         s->id = user_id;
         HASH_ADD_INT( users, id, s );  /* id: name of key field */
     }
+    assert(name != NULL && strlen(name) < sizeof(s->name));
     strcpy(s->name, name);
 }
 
@@ -77,9 +81,24 @@ void sort_by_id()
     HASH_SORT(users, id_sort);
 }
 
+char *get_input(char *buf, int siz, FILE *stream)
+{
+    char * ret = fgets(buf, siz, stream);
+
+    if (ret != NULL) {
+        size_t len = strlen(buf);
+        if (len > 0 && buf[len - 1] == '\n')
+        {
+            buf[len - 1] = '\0';
+        }
+    }
+
+    return ret;
+}
+
 int main()
 {
-    char in[10];
+    char in[NAME_MAXLEN + 1];
     int id=1, running=1;
     struct my_struct *s;
     unsigned num_users;
@@ -95,27 +114,27 @@ int main()
         printf(" 8. print users\n");
         printf(" 9. count users\n");
         printf("10. quit\n");
-        gets(in);
+        get_input(in, sizeof(in), stdin);
         switch(atoi(in)) {
             case 1:
                 printf("name?\n");
-                add_user(id++, gets(in));
+                add_user(id++, get_input(in, sizeof(in), stdin));
                 break;
             case 2:
                 printf("id?\n");
-                gets(in);
+                get_input(in, sizeof(in), stdin);
                 id = atoi(in);
                 printf("name?\n");
-                add_user(id, gets(in));
+                add_user(id, get_input(in, sizeof(in), stdin));
                 break;
             case 3:
                 printf("id?\n");
-                s = find_user(atoi(gets(in)));
+                s = find_user(atoi(get_input(in, sizeof(in), stdin)));
                 printf("user: %s\n", s ? s->name : "unknown");
                 break;
             case 4:
                 printf("id?\n");
-                s = find_user(atoi(gets(in)));
+                s = find_user(atoi(get_input(in, sizeof(in), stdin)));
                 if (s) {
                     delete_user(s);
                 } else {


tests/example.c
```diff
@@ -21,7 +21,7 @@ void add_user(int user_id, char *name)
         s->id = user_id;
         HASH_ADD_INT( users, id, s );  /* id: name of key field */
     }
-    strcpy(s->name, name);
+    strncpy(s->name, name, sizeof(s->name));
```
Quuxplusone, line 24: `strcpy` is correct — or at least, it doesn't hide the bad behavior if `name` is too long for the `s->name` field.

`strncpy` would avoid the buffer overflow on long names, but only by truncating the entered `name` — and incidentally, if the `name` were shorter than `sizeof(s->name)`, it'd pad with null bytes (which is harmless but inefficient).

The way to make this code "safe" would be to slap an `assert(strlen(name) < sizeof(s->name))` on the line above... but as this is just a toy example, I don't know that that's warranted, either.

tests/example.c
```diff
@@ -95,27 +95,27 @@ int main()
...
+                add_user(id++, fgets(in, sizeof(in), stdin));
                 break;
             case 2:
                 printf("id?\n");
-                gets(in);
+                fgets(in, sizeof(in), stdin);
                 id = atoi(in);
                 printf("name?\n");
-                add_user(id, gets(in));
+                add_user(id, fgets(in, sizeof(in), stdin));
```
Quuxplusone, line 109: I do like the idea of eliminating `gets` here — not so much for the buffer-overflow issue, as because it is no longer standard C. (It was removed in C11.) However:

Just as `fputs`' behavior is different from `puts`', `fgets`' behavior is different from `gets`'. Specifically, `gets` trims the trailing newline, but `fgets` does not. (Autobiographical note: [I once wrote some code](http://www.club.cc.cmu.edu/~ajo/free-software/getline/getline.html) to work around the quirks here. But I'm not pulling in that library just for this toy example!)

So this patch is changing behavior — if the user enters `Arthur`, it'll be stored in the hash as `"Arthur\n"`. Which is... fine, I guess... but it feels wrong. Could we perhaps factor out two helper functions `int get_int()` and `const char *get_name(char *buf, int n)`, and use them throughout? How much uglier would that make the code?

tests/example.c
```diff
@@ -95,27 +95,27 @@ int main()
...
             case 3:
                 printf("id?\n");
-                s = find_user(atoi(gets(in)));
+                s = find_user(atoi(fgets(in, sizeof(in), stdin)));
                 printf("user: %s\n", s ? s->name : "unknown");
                 break;
             case 4:
                 printf("id?\n");
-                s = find_user(atoi(gets(in)));
+                s = find_user(atoi(fgets(in, sizeof(in), stdin)));
```
Quuxplusone, line 118: P.S.: Whatever changes we make to `example.c` also need to be reflected in `doc/userguide.txt`.

Quuxplusone (changes requested):

tests/example.c
lnslbrty: > `strcpy` is correct — or at least, it doesn't hide the bad behavior if `name` is too long for the `s->name` field.

It is neither correct nor does it give the user any useful information why it just crashed. And leaving an exploitable buffer overflow there for cosmetic reasons is not a sign of code quality.

> `strncpy` would avoid the buffer overflow on long names, but only by truncating the entered `name` — and incidentally, if the `name` were shorter than `sizeof(s->name)`, it'd pad with null bytes (which is harmless but inefficient).
> 
> The way to make this code "safe" would be to slap an `assert(strlen(name) < sizeof(s->name))` on the line above... but as this is just a toy example, I don't know that that's warranted, either.

I would not use assertions in that situation but it gives at least some information why this application stops doing what it should do.

tests/example.c
lnslbrty: > I do like the idea of eliminating `gets` here — not so much for the buffer-overflow issue, as because it is no longer standard C. (It was removed in C11.) However:
> 
> Just as `fputs`' behavior is different from `puts`', `fgets`' behavior is different from `gets`'. Specifically, `gets` trims the trailing newline, but `fgets` does not. (Autobiographical note: [I once wrote some code](http://www.club.cc.cmu.edu/~ajo/free-software/getline/getline.html) to work around the quirks here. But I'm not pulling in that library just for this toy example!)

Removing a newline can be done with less code then described in  [I once wrote some code](http://www.club.cc.cmu.edu/~ajo/free-software/getline/getline.html).
> 
> So this patch is changing behavior — if the user enters `Arthur`, it'll be stored in the hash as `"Arthur\n"`. Which is... fine, I guess... but it feels wrong. Could we perhaps factor out two helper functions `int get_int()` and `const char *get_name(char *buf, int n)`, and use them throughout? How much uglier would that make the code?

I agree, will fix that.

tests/example.c
lnslbrty: If the patch is finished, there won't be any userguide changes necessary anymore.

tests/example.c
lnslbrty: > `strncpy` would avoid the buffer overflow on long names, but only by truncating the entered `name` — and incidentally, if the `name` were shorter than `sizeof(s->name)`, it'd pad with null bytes (which is harmless but inefficient).

By the way: There is no NUL padding, just one trailing NUL byte. See `man 3 strncpy`.

Quuxplusone: Today you learned. :) https://linux.die.net/man/3/strncpy

lnslbrty: Agreed. At least one useful outcome from this discussion.

pdebuyl:
Hi @Quuxplusone it might be appropriate to close this PR as gets has been removed.

Quuxplusone:
I agree, this was sufficiently fixed by a109c6b702b42cb5e134f489bb48a2b5aaa88ade.

PR #219, opened 2021-01-02
cleanup whitespaces

mwarning:
Feel free to merge or fold into other commits. :-)

### DIFF
--- a/doc/index.html
+++ b/doc/index.html
@@ -72,7 +72,7 @@ struct my_struct {
 struct my_struct *users = NULL;
 
 void add_user(struct my_struct *s) {
-    HASH_ADD_INT( users, id, s );
+    HASH_ADD_INT(users, id, s);
 }
 
 </pre>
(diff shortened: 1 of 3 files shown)


mwarning:
I am not sure macro call arguments should have a whitespace padding. But since it is not consistent, I like to keep it the same as the normal function call syntax.

Quuxplusone:
Seems harmless enough to me.

Irrelevantly to this PR, but just FYI, in "uthash.h" itself we have an idiosyncratic style where *macro definitions* (alone) lack spaces. Troy was doing this consistently before I started contributing, I've kept it up for no particular reason other than consistency, and I'd probably reject a PR to "fix" it. For example:

    #define HASH_ADD_INORDER(hh,head,fieldname,keylen_in,add,cmpfcn)                 \
      HASH_ADD_KEYPTR_INORDER(hh, head, &((add)->fieldname), keylen_in, add, cmpfcn)

mwarning:
Thanks for the context.

Quuxplusone:
Merged 524ca1a03bdab388fcbdb766383d6c02b5f7d0e1 and 37d20214271302895d4d24fbe727f2b690ecb6e5 — thanks!

PR #224, opened 2021-01-27
Reset the i in utstring_done, utringbuffer_done and utarray_done

gvcat:
This change allows us to reuse `a` without calling the initialization method.


Quuxplusone:
I haven't looked closely yet, but my kneejerk reaction is that `utstring_done` is like a "destructor," i.e.

- if you're trying to clean up a string _that you plan to reuse immediately as an empty string_, you should be calling `utstring_clear` instead of `utstring_done`, so that you can reuse its existing capacity

- `utstring_done` shouldn't be zeroing out `(s)->n` either

But I should look at the git log and see whether there was some subtle reason for `utstring_done` to be zeroing out `(s)->n`. At least I agree that _if_ it zeroes out `n`, it should definitely also zero out `i`.

edwin-oetelaar:
Can you present a situation where modification of existing code solves a known bug or improves performance.
What problem is solved by changing the code?
Please show reason for change.

robotic-cat:
Sorry I missed something earlier, I pushed the new changes.

If `utstring_done` can reset `s`, we can call `utstring_bincpy` and other methods directly without calling `utstring_init`. `utstring_done` can free up previous memory while we can still use `s`, and if the application is memory sensitive calling `utstring_done` can save a lot of memory space.

After we call `utstring_done`, `utstring_len` and `utstring_body` may still be used, and it would be more appropriate if these two methods returned `0` and `NULL`.

Quuxplusone:
> After we call `utstring_done`, `utstring_len` and `utstring_body` may still be used

Well, that's the thing — they can't! `utstring_done` says that you're _done_ with the string. It's not supposed to be a substitute for `utstring_clear`.

What's the situation where you can't either
- use `utstring_clear` to clear the string (without heap traffic), or
- use `utstring_done` followed by `utstring_init` to destroy and reinitialize the string (freeing the heap allocation)
?

PR #227, opened 2021-03-01
Replace *.github.com urls with *.github.io

troydhanson:
GitHub is ending the *.github.com redirects in April. Use *.github.io instead.


Quuxplusone:
I see two more instances of `git grep troydhanson.github.com` that you might as well fix at the same time. And then LGTM, ship it!

PR #230, opened 2021-05-14
Fix typo in userguide

ylecuyer:
Hello,

I have been reading the user guide and I found a typo in this example.
Otherwise `name` is not defined


Quuxplusone:
`HASH_ADD_STR` accepts a field name (not a C expression) as its second argument, so I believe this is correct as written.
Compare some of the uses of `HASH_ADD_STR` in the unit tests:
```
test59.c:    HASH_ADD_STR(items, name, i);
test59.c:    HASH_ADD_STR(i->sub, name, s);
test60.c:    HASH_ADD_STR(items, name, i);
test60.c:    HASH_ADD_STR(i->sub, name, s);
```

PR #234, opened 2021-07-08
Fix -Wimplicit-fallthrough warnings

mwileczka:
The /* FALLTHROUGH */ comment must be on a new line from case statement, otherwise the compiler doesn't detect it.

https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough

### DIFF
--- a/src/uthash.h
+++ b/src/uthash.h
@@ -668,17 +668,28 @@ do {                                                                            
   }                                                                              \
   hashv += (unsigned)(keylen);                                                   \
   switch ( _hj_k ) {                                                             \
-    case 11: hashv += ( (unsigned)_hj_key[10] << 24 ); /* FALLTHROUGH */         \
-    case 10: hashv += ( (unsigned)_hj_key[9] << 16 );  /* FALLTHROUGH */         \
-    case 9:  hashv += ( (unsigned)_hj_key[8] << 8 );   /* FALLTHROUGH */         \
-    case 8:  _hj_j += ( (unsigned)_hj_key[7] << 24 );  /* FALLTHROUGH */         \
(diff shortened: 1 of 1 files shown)


Quuxplusone:
The GCC documentation to which you linked doesn't say anything about "must be on a new line."
`-Wimplicit-fallthrough=` was added in GCC 7. If you're using GCC, there's no problem with the current code:
https://godbolt.org/z/PjEYEEcoz

Do you need this reformatting for some non-GCC compiler? Which compiler?

mwileczka:
Sorry if I jumped the gun. I'm using Ubuntu 18.04 with gcc 7.5.0.
We have a lot gcc warning flags turned on.
I was unable to replicate my problem with the included test files, even by adding all of the compiler flags we use.

The only way I can bypass the warnings in our projects is to use -Wimplicit-fallthrough=0 or move the /* FALLTHROUGH */ to the next line.

I'll try to replicate the problem before I repost.

Thank you for all of your hard work - we use this package extensively.

mwileczka:
I found out that using `-save-temps=obj` caused the warnings to show up.
I'll closed this out.

Quuxplusone:
@mwileczka: With my uthash-maintainer hat on, I approve. :)  With my hat off, though, that's super weird, isn't it? Why should `-save-temps=obj` cause extra warnings... and _especially_ why should it cause warnings that are whitespace-dependent?
I could imagine that somehow you're preprocessing the file and causing it to eat all the comments, including the significant `FALLTHROUGH` comments. But I can't see how putting the comment on its own line would have fixed that, either.

Issue #243, opened 2022-06-25
Replace TravisCI with GitHub Actions or AppVeyor?


Quuxplusone:
In June 2021, TravisCI changed its domain from travis-ci.org to travis-ci.com, added "plans" and "credits," and generally seems to have made it a non-trivial process to get uthash's TravisCI build working again. I'm also worried about sustainability — if we get something working again, will it just deplete Troy's "credits" and then stop working again?
https://app.travis-ci.com/github/troydhanson/uthash

So we (which is to say, I eventually, or hopefully some friendly contributor before then) should probably look into some other way to do CI testing on pull requests and periodically on `master`. GitHub Actions is supposed to offer something like that; alternatively, AppVeyor still seems to be free, and offers Linux builds now, so maybe we can switch to AppVeyor.

PR #245, opened 2022-08-14
ci: add GitHub Actions CI

aloisklink:
Adds a basic GitHub Actions CI action that copies https://github.com/troydhanson/uthash/blob/master/.travis.yml

Fixes https://github.com/troydhanson/uthash/issues/243

This action will run on:
  - a `git push` to any branch on any fork.
  - a PR just to the `master` branch

I've purposely kept the `.travis.yml` file, since we can always just use both GitHub Actions and Travis for a while, before deleting Travis once the maintainers are comfortable with GitHub Actions CI.

Potential improvements:
  - We could use https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs to reduce some of the copy-and-paste, but I couldn't think of any easy way to put both Clang/GCC into a Matrix.
  - We could add a GitHub Action that automatically builds the documentation, and publishes it to GitHub pages.
  - We can also add a CI badge to the README.md file:
    [Build and Test](https://github.com/troydhanson/uthash/actions/workflows/build.yml)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,8 +6,11 @@ on:
     branches: [ "master" ]
 
 jobs:
-  build:
-    runs-on: ubuntu-latest
+  build-gcc:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v3
     - run: make -C tests EXTRA_CFLAGS="-W -Wall -Wextra -Wswitch-default"
@@ -16,22 +19,13 @@ jobs:
     - run: make -C tests clean ; make -C tests cplusplus
     - run: make -C tests clean ; make -C tests cplusplus EXTRA_CFLAGS=-DNO_DECLTYPE
   build-clang:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
     env:
       CC: clang
       CXX: clang++
-    steps:
-    - uses: actions/checkout@v3
-    - run: make -C tests EXTRA_CFLAGS="-W -Wall -Wextra -Wswitch-default"
-    - run: make -C tests clean ; make -C tests pedantic
-    - run: make -C tests clean ; make -C tests pedantic EXTRA_CFLAGS=-DNO_DECLTYPE
-    - run: make -C tests clean ; make -C tests cplusplus
-    - run: make -C tests clean ; make -C tests cplusplus EXTRA_CFLAGS=-DNO_DECLTYPE
-  build-mac:
-    runs-on: macos-latest
-    needs:
-      - build
-      - build-clang
     steps:
     - uses: actions/checkout@v3
     - run: make -C tests EXTRA_CFLAGS="-W -Wall -Wextra -Wswitch-default"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,41 @@
+name: Build and Test
+
+on:
+  push: # any branch
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - run: make -C tests EXTRA_CFLAGS="-W -Wall -Wextra -Wswitch-default"
+    - run: make -C tests clean ; make -C tests pedantic
+    - run: make -C tests clean ; make -C tests pedantic EXTRA_CFLAGS=-DNO_DECLTYPE
+    - run: make -C tests clean ; make -C tests cplusplus
+    - run: make -C tests clean ; make -C tests cplusplus EXTRA_CFLAGS=-DNO_DECLTYPE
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,41 @@
...
+    - run: make -C tests EXTRA_CFLAGS="-W -Wall -Wextra -Wswitch-default"
+    - run: make -C tests clean ; make -C tests pedantic
+    - run: make -C tests clean ; make -C tests pedantic EXTRA_CFLAGS=-DNO_DECLTYPE
+    - run: make -C tests clean ; make -C tests cplusplus
+    - run: make -C tests clean ; make -C tests cplusplus EXTRA_CFLAGS=-DNO_DECLTYPE
+  build-mac:
+    runs-on: macos-latest
+    needs:
+      - build
+      - build-clang
```
aloisklink, line 34: I've set the MacOS CI to only run if the Linux CIs pass, since even though GitHub Actions is free for open-source repos, there might be hidden soft caps (or maybe they'll introduce caps in the future).

Normally MacOS CI costs 10x the Linux CIs on GitHub Actions.

Quuxplusone: That makes sense. Or maybe gate it only on `build-clang`, since Mac uses Clang. (But I believe I see how to make that change myself — just delete one line, right? :))

Naming nit: let's rename `build` to `build-gcc`, if I understand it correctly.

Quuxplusone:
> reduce some of the copy-and-paste

I don't think that's worth it; I'm _very_ comfortable with copy-paste and diff, and not nearly so much with random API minilanguages. :)

> We could add a GitHub Action that automatically builds the documentation, and publishes it to GitHub pages.

I don't think that's worth it. Or at least not the auto-publishing part. Having a CI job that makes sure `asciidoc` still runs successfully, would probably be nice.

> We can also add a CI badge to the README.md file

Yes please. :)

Does GitHub support Windows builds too? Could we do a Windows/MSVC build this same way?

aloisklink:
I've renamed the builds into `build-gcc` and `build-clang` in 5ddf4ef3133c2df881044673ae257a46288e7a33

I've also merged the MacOS builds into them since:
  - It's pretty easy to do (it's a random CI mini-language, but at least picking OSes is fairly straight forward and common (checking for Clang/GCC would have been harder would have been a lot more complicated))
  - It's got some nicer UIs
  - Even if MacOS costs 10x the Linux CI runners, `uthash` is quite small, so the CI only takes a minute or two, so it should be a lot faster than the multi-hour builds that some other projects have.

> Having a CI job that makes sure asciidoc still runs successfully, would probably be nice.

Awesome! I'll make another PR then to keep things separate.

> We can also add a CI badge to the README.md file

:+1: I'd recommend we only look into doing this after this gets merged.
It should just be copying and pasting the following, but as it currently says "No build found", we probably need to wait until this gets merged to confirm it works:

```md
[![Build and Test](https://github.com/troydhanson/uthash/actions/workflows/build.yml/badge.svg)](https://github.com/troydhanson/uthash/actions/workflows/build.yml)
```

> Does GitHub support Windows builds too? Could we do a Windows/MSVC build this same way?
Yes, the [`windows-latest` image has MSVC](https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md#microsoft-visual-c)

I did try a basic Windows CI test and got the following error, see commit https://github.com/aloisklink/uthash/commit/3a587e3c6a554aff253a72cf5c6999f6b921cb9e and CI logs https://github.com/aloisklink/uthash/runs/7847877279?check_suite_focus=true:

```console
> make -C tests EXTRA_CFLAGS="-W -Wall -Wextra -Wswitch-default"
mingw32-make: Entering directory 'D:/a/uthash/uthash/tests'
cc  -I../src -g -Wall -W -Wall -Wextra -Wswitch-default  -o test1 test1.c
process_begin: CreateProcess(NULL, cc -I../src -g -Wall -W -Wall -Wextra -Wswitch-default -o test1 test1.c, ...) failed.
make (e=2): The system cannot find the file specified.
mingw32-make: Leaving directory 'D:/a/uthash/uthash/tests'

mingw32-make: *** [makefile:[8](https://github.com/aloisklink/uthash/runs/7847877279?check_suite_focus=true#step:3:9)8: test1] Error 2
```

Unfortunately, it's been many years since I last had the pleasure of working with Windows, so I've got no idea what the issue could be. My gut feeling is that maybe we need to use CMake instead of GNU Make to write the tests for Windows support, but again, my Windows knowledge is very out of date.

.github/workflows/build.yml
aloisklink: Renamed to `build-gcc` and `build-clang` in 5ddf4ef3133c2df881044673ae257a46288e7a33

I've also combined the MacOS builds into `build-gcc` and `build-clang`, since even though it's a random API minilanguage, at least this one is pretty basic.

(plus, it's probably overkill to try to reduce our MacOS CI usage. This repo's CI takes only 1-2 minutes, and other projects have hours worth of CI time)

And, the UI is pretty nice if you use the `matrix` feature:

image

Quuxplusone:
Merged 1e0baf069b24852839102fbae43b8c23c6f26a2a!

PR #247, opened 2022-09-24
Update uthash.h and utlist.h to compile with the elbrus C compiler.

utf-4096:
This fixes uthash.h and utlist.h to compile as expected with LCC (eLbrus C Compiler).
Elbrus has the `__LCC__` macro which is named the same as LCC (Retargetable Compiler for ANSI C), however, Elbrus supports `__typeof()`, so this adds a check for `__e2k__` which is defined by it.

### DIFF
--- a/src/uthash.h
+++ b/src/uthash.h
@@ -51,7 +51,7 @@ typedef unsigned char uint8_t;
 #else                   /* VS2008 or older (or VS2010 in C mode) */
 #define NO_DECLTYPE
 #endif
-#elif defined(__BORLANDC__) || defined(__ICCARM__) || defined(__LCC__) && !defined(__e2k__) || defined(__WATCOMC__)
+#elif defined(__BORLANDC__) || defined(__ICCARM__) || defined(__LCC__) && !defined(__MCST__) || defined(__WATCOMC__)
 #define NO_DECLTYPE
 #else                   /* GNU, Sun and other compilers */
 #define DECLTYPE(x) (__typeof(x))
--- a/src/uthash.h
+++ b/src/uthash.h
@@ -51,7 +51,7 @@ typedef unsigned char uint8_t;
 #else                   /* VS2008 or older (or VS2010 in C mode) */
 #define NO_DECLTYPE
 #endif
-#elif defined(__BORLANDC__) || defined(__ICCARM__) || defined(__LCC__) || defined(__WATCOMC__)
+#elif defined(__BORLANDC__) || defined(__ICCARM__) || defined(__LCC__) && !defined(__e2k__) || defined(__WATCOMC__)
 #define NO_DECLTYPE
 #else                   /* GNU, Sun and other compilers */
 #define DECLTYPE(x) (__typeof(x))
--- a/src/utlist.h
+++ b/src/utlist.h
@@ -70,7 +70,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #else                   /* VS2008 or older (or VS2010 in C mode) */
 #define NO_DECLTYPE
 #endif
-#elif defined(__BORLANDC__) || defined(__ICCARM__) || defined(__LCC__) && !defined(__e2k__) || defined(__WATCOMC__)
+#elif defined(__BORLANDC__) || defined(__ICCARM__) || defined(__LCC__) && !defined(__MCST__) || defined(__WATCOMC__)
 #define NO_DECLTYPE
 #else                   /* GNU, Sun and other compilers */
 #define LDECLTYPE(x) __typeof(x)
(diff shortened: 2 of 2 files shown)


Quuxplusone:
This looks good to me (except that I'll add some parentheses to emphasize the precedence before merging). However, can I ask:

(1) What was the symptom you were seeing on Elbrus? It seems like you're saying that Elbrus had been using the `NO_DECLTYPE` codepaths and those codepaths _didn't work_ for you? What error messages did you see from the compiler, for what kind of source code? (I.e., is this patch working around a deeper problem with `NO_DECLTYPE`, such that we might end up fixing and/or unsupporting that codepath entirely?)

(2) Is it possible to test Elbrus anywhere online (e.g. [Godbolt Compiler Explorer](https://godbolt.org/)), or otherwise get access to it?

(3) Upstream comment: It seems that https://drh.github.io/lcc/ already uses the name `__LCC__`, and also defines hardware–related macros such as `__sparc__`, `__i386__`, `__alpha`. _Theoretically_, if LCC were ever "retargeted" to support Elbrus 2000 hardware, it would _also_ define both `__LCC__` and `__eb2k__`, right? Wouldn't it be a fantastic idea for the upstream Elbrus C Compiler people to adopt a set of macros that clearly and _unambiguously_ distinguished "Elbrus C Compiler" from LCC?

utf-4096:
(1) NO_DECLTYPE codepaths were giving this error while trying to compile [SpadesX/SpadesX](https://github.com/SpadesX/SpadesX):
```c
lcc: "/export/home/utf/spadesx/SpadesX/Source/Util/Utlist.h", line 144: error #3263:
          argument of type "char *" is incompatible with parameter of type
          "Command *" [-Werror=incompatible-pointer-types]
            } else if (cmp(_ls_p,_ls_q) <= 0) {                                                  \
                           ^
 in expansion of macro "LL_SORT2" at line 109
 in expansion of macro "LL_SORT" at line 934 of
           "/export/home/utf/spadesx/SpadesX/Source/Commands.c"
      LL_SORT(server->commandsList, commandCompare);
      ^

lcc: "/export/home/utf/spadesx/SpadesX/Source/Util/Utlist.h", line 144: error #3263:
          argument of type "char *" is incompatible with parameter of type
          "Command *" [-Werror=incompatible-pointer-types]
            } else if (cmp(_ls_p,_ls_q) <= 0) {                                                  \
                                 ^
 in expansion of macro "LL_SORT2" at line 109
 in expansion of macro "LL_SORT" at line 934 of
           "/export/home/utf/spadesx/SpadesX/Source/Commands.c"
      LL_SORT(server->commandsList, commandCompare);
      ^

2 errors detected in the compilation of "/export/home/utf/spadesx/SpadesX/Source/Commands.c".
```

(2) It has a CE at https://ce.mentality.rip/ with different compiler versions.
(3) I've now talked to people who have more knowledge around MCST LCC, and it has its own macro, `__MCST__`  (for compiler v1.25+). I'll update my code to check for this instead. Also, I've been informed there is an E2K version of drh's lcc currently being worked on (though, it's currently private).

Quuxplusone:
Note to self — it might even be time to flip this preprocessor stuff inside-out:
```
#if defined(DECLTYPE) || defined(NO_DECLTYPE)
/* do nothing */
#elif defined(__cplusplus) && (_MSC_VER >= 1600)
#define DECLTYPE(x) (decltype(x))
#elif defined(__cplusplus) && (__cpp_decltype >= 200707L))
#define DECLTYPE(x) (decltype(x))
#elif defined(__MCST__)
#define DECLTYPE(x) (__typeof(x))
#elif defined(__BORLANDC__) || defined(__ICCARM__) || defined(__LCC__) || defined(__WATCOMC__)
#define NO_DECLTYPE
#else /* the default for GNU, Sun and other compilers */
#define DECLTYPE(x) (__typeof(x))
#endif
```
It's unfortunate that I don't know what compilers people are still using today that actually hit the "else" branch, and all the macros defined by the `NO_DECLTYPE` compilers. Like, could we add
```
#elif defined(__MCST__) || defined (__GNUC__)
#define DECLTYPE(x) (__typeof(x))
```
at this point? or would that break some ancient GCC and/or some recent non-GNU-compiler-that-nonetheless-defines-`__GNUC__`? I don't think we have any real way of knowing, other than "try it and see who complains."

Quuxplusone:
> There is a list of compilers and their macros at https://github.com/cpredef/predef/blob/master/Compilers.md which seems pretty long and accurate.

Yes, but (1) that list doesn't include many of the compilers checked-for in uthash today (e.g. Borland C++, Watcom C++), (2) it doesn't give any hint as to which ones support `__typeof`, (3) it doesn't give any hint as to which _versions_ of those compilers support `__typeof` (e.g. today's code implies that GCC pre-4.3 didn't support it), (4) it doesn't correlate with which compilers might actually be in use by uthash's users today. Sadly the people I most don't want to break, are the people who have never interacted with the project because it Just Worked on their system. 😛 

Anyway, I've asked StackOverflow here:
https://stackoverflow.com/questions/73848714/detect-if-a-c-compiler-supports-typeof-and-or-decltype
and I'll go ahead and merge this patch (with some editorial tweaks) now.

Quuxplusone:
Merged 4d01591ed9b67ff153410d0f0cfa47f51b735eb4!

PR #246, opened 2022-08-16
ci: add GitHub Action to build docs

aloisklink:
Adds a basic GitHub Action that builds the documentation.

As discussed in https://github.com/troydhanson/uthash/pull/245#issuecomment-1215249856, I haven't added automatic publishing to GitHub Pages.

However, if you do want automatic publishing of docs to GitHub pages, GitHub has recently added the feature where you can automatically publish a website through GitHub Actions, see <https://github.blog/changelog/2022-07-27-github-pages-custom-github-actions-workflows-beta/>:

It's still in Beta, but all you'd have to add is the following:

```yaml
      # old contents ...
      - name: installs docs to /tmp/uthash-gh-pages
        run: make -C doc stage
      - name: Upload docs
        uses: actions/upload-pages-artifact@v1
        with:
          path: /tmp/uthash-gh-pages
  publish:
    name: publish docs (main-branch only)
    if: github.ref == 'refs/heads/main'
    needs: build-docs
    # Grant GITHUB_TOKEN the permissions required to make a Pages deployment
    permissions:
      pages: write      # to deploy to Pages
      id-token: write   # to verify the deployment originates from an appropriate source
    # Deploy to the github-pages environment
    environment:
      name: github-pages
      url: ${{ steps.deployment.outputs.page_url }}
    # Specify runner + deployment step
    runs-on: ubuntu-latest
    steps:
      - name: Deploy to GitHub Pages
        id: deployment
        uses: actions/deploy-pages@v1
```

PR #248, opened 2022-09-26
GitHub CI: Also test building the docs

Quuxplusone:
Basically #246, but with fewer files.

@aloisklink, what's the advantage of having two separate .yml files here? IIUC, the (only) advantage is that it would let us have two different README badges, e.g. "build is green but docs are red" — but actually I'd never want to get into that state, so, I think this single-build.yml approach makes more sense; am I right?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: Build and Test
+name: build # This name shows up in badge.svg
 
 on:
   push: # any branch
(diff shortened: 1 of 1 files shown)


aloisklink:
> I think this single build.yml approach makes more sense; am I right?

If the docs and build/test/release don't overlap much, most projects keep things separate since it some advantages. But if you don't think you'll use these features, you could keep a single `build.yml` file if you prefer it.

---
**One minor benefit** is secrets/permissions/env variables. Again, this project doesn't use them, but separate workflows make it easier to keep [permissions](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#overview) separate.

---
**One minor benefit** is that you can disable actions on a per-`.yml`-file level if the docs builds breaks due to a minor temporary issue. E.g. if `sudo apt install asciidoc` fails because of a bug, and you're waiting for Ubuntu to release an asciidoc bugfix.

(In the early days of GitHub Actions, `sudo apt install ...` would often fail for a few hours, when a package was updated, but GitHub/Microsoft's cache didn't update)

image

---
**Another minor benefit** is that it lets you change the `on: ...` stuff later easily:

https://github.com/troydhanson/uthash/blob/83bb057cda7d885490f547a889545c618d77ebb1/.github/workflows/build.yml#L3-L6

For larger projects (especially if you're paying for CI!), you'd often want the `build.yml` file to have something like the following, so the build CI doesn't run for commits/PRs that only modify documentation.

```yaml
on:
  push:
    paths-ignore:
      - docs/**
```

Quuxplusone:
Merged 1a53f304aa2db47e8584ddcb16a6587578e245f3

PR #251, opened 2022-11-07
utarray: utarray_oom on strdup failure, don't use strdup, and remove `-Doom` backward-compat hack

Quuxplusone:


aloisklink (approved):
Looks great to me!

> We changed `-Doom=...` to `-Dutarray_oom=...` in commit https://github.com/Quuxplusone/uthash/commit/cfc95dbe14b8aa47cee0292158876dd118903e48,
part of the uthash 2.1.0 release (Dec 2018). Since then, we've had
uthash 2.2.0 (Dec 2020) and 2.3.0 (Feb 2021). I think that's enough
time for people to update their projects. Remove the backward-compatibility
hack `#define utarray_oom oom`.

I had a quick look at https://pkgs.org/search/?q=uthash

It looks like every OS has had a major version that included uthash v2.1.0 or later, except for Solus, which is still on v2.0.1.

I've never heard of Solus, so I don't really know how popular they are, and whether it's worth holding back on that change (it looks like there is some development on their `uthash` port): https://dev.getsol.us/source/uthash

PR #260, opened 2023-12-10
Follow commenting convention for copyright headers.

jkoshy:
Use "/*-" for the first line of copyright header comments.

Doing so allows automation to collect license information in a source tree by looking for comments that start with "/*-".

This is a comment-only change that does not affect code behavior.


Quuxplusone:
This seems reasonable enough, but the commit message should explain something about why we're doing this and why this (specifically) is the right thing to do. You say this "allows automation to ..."; but _whose_ automation? is there a well-known standard for this? do any existing tools support special behavior for `/*-` comments?
A quick google for "copyright header comments" didn't turn up any mentions of `/*-`.

jkoshy:
Thanks for the review.   I have added more context to the commit message.

jkoshy:
> Is anyone willing to say "yes, this change will specifically make my
> workflow easier, in the following concrete way..."

The change would make my own workflow a tiny bit easier.  For example, a change like [[Elftoolchain:r4024](https://sourceforge.net/p/elftoolchain/code/4024/)] could become a copy instead of a copy-edit.

This convenience propagates downstream as `uthash`'s source code ends up in downstream projects via imports of Elftoolchain code&mdash;for example: [FreeBSD](https://cgit.freebsd.org/src/tree/contrib/elftoolchain/common), [Minix](https://github.com/Stichting-MINIX-Research-Foundation/minix/tree/master/external/bsd/elftoolchain/dist/common), [NetBSD](http://cvsweb.netbsd.org/bsdweb.cgi/src/external/bsd/elftoolchain/dist/common/), [RTEMS](https://github.com/RTEMS/rtems-tools/tree/master/rtemstoolkit/elftoolchain/common).

Following the `/*-` convention upstream would make it easier for users of *those* open-source projects to locate and process the (c) blocks in their source base correctly.

> who specifically will be helped by it?

1.  Developers using `uthash`, either directly or indirectly (as above), who need to conduct due diligence around the licenses in their source trees.
2. People wishing to preserve pre-formatted content such as a (c) coment block when running formatters over their C/C++ code.

> Or to put it another way, suppose we did this, and then
> next year someone comes and asks for /*- to be replaced
> with /***

Although possible, this seems unlikely because tooling support for the alternate `/***` marker would also need to be implemented.

BSD projects have used  `/*-` to mark (c) comments for over [three decades](https://github.com/weiss/original-bsd/blame/master/lib/libc/vax/string/memset.s), as of the writing.  *BSD tools that process source code (for example, [BSD indent(1)](https://man.freebsd.org/cgi/man.cgi?query=indent)) know about the `/*-` convention.  Such tooling would need to be augmented to support the alternate `/***` convention, which would be a lot of work for (IMHO) little value.

> should we also be using a SPDX-License-Identifier: line

The SPDX identifier has a different function than the `/*-` marker.  IMHO it would be useful to add it.

Would you like me to revise the patch to add in the appropriate SPDX line?

Quuxplusone:
> People wishing to preserve pre-formatted content such as a (c) comment block when running formatters over their C/C++ code.

That's a good point. `clang-format` is smart enough to leave block comments alone, but as you (partly) pointed out, BSD `indent` is not. [`indent`](https://man.freebsd.org/cgi/man.cgi?query=indent) knows about "box comments" of the form `[/][*][*-]+`. But `indent` doesn't _intrinsically_ treat `/*-` any differently from, well, `/***`! (Verified via my MacBook's `/usr/bin/indent`.)

> The change would make my own workflow a tiny bit easier. For example, a change like [[Elftoolchain:r4024](https://sourceforge.net/p/elftoolchain/code/4024/)] could become a copy instead of a copy-edit.

OTOH (moving the goalposts, I admit), that edit was needed only to preserve [[r3983]](https://sourceforge.net/p/elftoolchain/code/3983/) (May 2022)'s addition of `/*-`.

> The SPDX identifier has a different function than the /*- marker. IMHO it would be useful to add it.
> Would you like me to revise the patch to add in the appropriate SPDX line?

I don't know! This highlights my basic concern here, that we're changing a (trivial) thing without a concrete notion that what we're changing it to is best, or even better. [[r4024]](https://sourceforge.net/p/elftoolchain/code/4024/) didn't bother to add a SPDX line, despite [the BSD "kernel source file style guide"](https://man.freebsd.org/cgi/man.cgi?query=style) mandating one; so why should we take the `/*-` mandate any more seriously than the SPDX mandate?

Another plausible option would be to _remove_ the copyright headers from all uthash's files and rely on [the existing `LICENSE` file](https://github.com/troydhanson/uthash/blob/master/LICENSE) to communicate that information. But that option, also, is not clearly better than the status quo and I'm not inclined to pursue it without a good reason.

The [Minix](https://github.com/Stichting-MINIX-Research-Foundation/minix/tree/master/external/bsd/elftoolchain/dist/common) repo you linked to uses plain `/*` in uthash.h (for at least the past 8 years), and I guess they've survived. [libpng](https://github.com/glennrp/libpng/blob/libpng16/png.h) uses plain `/*`. [zlib](https://github.com/madler/zlib/blob/develop/zlib.h) uses plain `/*`. [libc++](https://github.com/llvm/llvm-project/blob/main/libcxx/include/algorithm) uses plain `//` with `SPDX-License-Identifier`. [glibc](https://github.com/bminor/glibc/blob/master/inet/bug-if1.c) uses plain `/*`. [PCRE2](https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2_jit_misc.c) uses plain `/*`. [protobuf-c](https://github.com/protobuf-c/protobuf-c/blob/master/protobuf-c/protobuf-c.h) uses plain `/*`. [librdkafka](https://github.com/confluentinc/librdkafka/blob/master/src/rdkafka.h) uses plain `/*`. [xxHash](https://github.com/Cyan4973/xxHash/blob/dev/xxhash.h) uses plain `/*` (and uses `/*-****` for a few non-(c) comments).

I see no consensus in favor of `/*-` here, and while maybe there needs to be a "first penguin," I don't see why it should be uthash specifically.

PR #261, opened 2024-05-13
Fixed MSVC warning about conditional expression being constant

noodlecollie:
This is because the condition expands out to `1 != 0` if `HASH_BLOOM` is not defined. Instead, I've modified the `HASH_BLOOM_TEST` macro to include the `!= 0` when the left hand side will not just be `1`, and leave the comparison as `1` otherwise - apparently MSVC thinks `if ( 1 )` is fine.


Quuxplusone:
@noodlecollie thanks! I merged a slightly different version that should fix the MSVC warning equally well.

PR #264, opened 2024-10-04
Handling size more carefully.

lonicerae:
Should take care of the type size when doing malloc().


shantanugadgil:
if we want to be pedantic about size safety, shouldn't we be checking the size of `ssrc[0]` (unlikely, but just in case the data type of `ssrc` changes)

PR #269, opened 2025-02-27
added a new macro 'utarray_replace' to replace an element in array'

xunicatt:


Quuxplusone:
Right now this patch is (1) missing tests and documentation, and (2) missing a semicolon, so once you add a test using your new macro, that test will fail to compile.
I think we shouldn't merge this unless it's useful to someone (such as yourself). If it's useful, you must be using it. If you're using it, you must be able to provide some code that demonstrates its use. Put that code into a new test case. And then, make sure it compiles and does the right thing at runtime! Right now you're failing the "it compiles" check.

xunicatt:
Thanks for the response and clarification, I will open a new pull request once I fix those issues.

PR #272, opened 2025-04-11
Add utarray_replace macro to safely replace elements in UT_array

xunicatt:
This pull request introduces a new macro function `utarray_replace` in `utarray.h`, which enables safe replacement of an element in a `UT_array`.

Currently, modifying an element in a `UT_array` often involves retrieving a pointer via `utarray_eltptr` and then directly assigning a new value. However, this method bypasses any user-defined `ctor` or `dtor` functions, potentially leading to memory leaks or undefined behavior when managing complex data types.

The new `utarray_replace` macro ensures that when an element is replaced:

1. The existing element is properly destroyed using the `dtor`, if defined.
2.  The new element is constructed via the `ctor`, if defined.

Tests and Documentation has been added to explain usage and behavior of `utarray_replace`.


tests/test99.c
```diff
@@ -0,0 +1,24 @@
...
+
+int main(void) {
+    UT_array *nums;
+    int i, *p;
+
+    utarray_new(nums,&ut_int_icd);
+    for(i=0; i < 10; i++) utarray_push_back(nums,&i);
+
+    i = 110;
+    utarray_replace(nums, 5, &i);
```
Quuxplusone, line 12: One thing I'd be slightly worried about people misunderstanding about a "replace" API is that the `5` here is an index, not a value. To that end, I suggest that this test should initialize the array (on line 9) with values that don't match their indices. E.g.

    for (i = 1; i < 1024; i <<= 1) {
        utarray_push_back(nums, &i);
    }
    i = 110;
    utarray_replace(nums, 5, &i);

(Note the whitespace/brace style, while you're at it, please.)

tests/test99.c
```diff
@@ -0,0 +1,24 @@
...
+
+    for(p=(int*)utarray_front(nums);
+        p!=NULL;
+        p=(int*)utarray_next(nums,p)) {
+        printf("%d\n",*p);
+    }
+
+
+    utarray_free(nums);
+
```
Quuxplusone, line 22:
```suggestion
```

tests/test99.c
```diff
@@ -0,0 +1,24 @@
...
+    i = 110;
+    utarray_replace(nums, 5, &i);
+
+    for(p=(int*)utarray_front(nums);
+        p!=NULL;
+        p=(int*)utarray_next(nums,p)) {
+        printf("%d\n",*p);
+    }
+
```
Quuxplusone, line 19:
```suggestion
```

src/utarray.h
```diff
@@ -217,6 +217,17 @@ typedef struct {
   qsort((a)->d, (a)->i, (a)->icd.sz, cmp);                                    \
 } while(0)
 
+#define utarray_replace(a, j, ele) do {                                           \
```
Quuxplusone, line 220: Nit: please align the backslashes in the same column as all the others are aligned.

I suggest `ele` should be named `p` for consistency with `utarray_insert`. (Please also note the whitespace style and conform to it.)

I'm not sure about the argument order here. `utarray_insert(a,p,j)` has the new item _followed_ by the index. At first glance I'd assume that we should also have `utarray_replace(a,p,j)` — but is there a good reason you deviated here?

doc/utarray.txt
```diff
@@ -336,6 +336,7 @@ Operations
 | utarray_next(UT_array *a,void *e) | get element of a following e (front if e is NULL)
 | utarray_prev(UT_array *a,void *e) | get element of a before e (back if e is NULL)
 | utarray_back(UT_array *a) | get last element of a
+| utarray_replace(UT_array *a, int j, void *p) | replace an element at index j
```
Quuxplusone, line 339: Nit: This should go up next to `utarray_insert`.
(And see my comments below re the signature / argument order.)

src/utarray.h
```diff
@@ -217,6 +217,17 @@ typedef struct {
...
+  char *ptr = utarray_eltptr(a, j);                                               \
+  if (ptr) {                                                                      \
```
Quuxplusone, line 222: Why is this test here? Your (very short) documentation doesn't mention anything about only _conditionally_ replacing the element. If you want the semantics to be "Replace the given index, _unless_ that index is out of range, in which case this function is a no-op," then that definitely needs to be in your documentation, because it's not obvious at all.

At first glance I suggest removing this test: `j` should always refer to a valid element, so `ptr` should never be null, so the test can be removed. (That is, just like `utarray_erase`, this function has a _precondition_ that the inputs are valid.)

tests/test100.c
```diff
@@ -0,0 +1,53 @@
...
+
+void Data_init(void* _data) {
+    Data *data = (Data*)_data;
+    data->dept = (char*)malloc(sizeof(char) * 2);
+}
+
+void Data_deinit(void* _data) {
+    Data *data = (Data*)_data;
+    free(data->dept);
+}
```
Quuxplusone, line 19: Nit: no leading underscore please. I suggest `vdata`, or just `vp` or `item` or `elt`.
Nit: `sizeof(char)` is `1` by definition; please remove here and below.

This API _feels_ like you're splitting up the initialization of a `Data` in a weird way: you're allocating space for 2 characters here, but you don't actually get around to initializing them until after the return to `Data_init`'s caller. Normally, we would just say

    data->dept = strdup("A");

except that this "default constructor" function doesn't know the string `"A"`. But it's still allocating memory, and then the caller is doing the second half of the strdup via `memcpy`.

Is there a more realistic way to factor this code? Or, if this _is_ based on real code, can we use better naming or values to demonstrate why this way of factoring it makes sense?

tests/test100.c
```diff
@@ -0,0 +1,53 @@
...
+
+int main(void) {
+    UT_array *datas;
+    int i;
+    Data d, *dp;
+    char tmp[] = {'A', 0};
+
+    UT_icd data_icd = {sizeof(Data), Data_init, NULL, Data_deinit};
```
Quuxplusone, line 27: This smells wrong. You need also to give a `Data_copy` function pointer here, to avoid double-free bugs.

Now, `a->icd.copy` is called in only four places: `utarray_push_back`, `utarray_insert`, `utarray_replace`, and `utarray_inserta`. In the first three places, the source object is _outside_ utarray, controlled by the caller, and the caller can just choose to "copy" by memcpy and then remember not to "destroy" their own object. (That is, the caller can choose to pretend that `utarray_push_back` does _relocation_ instead of _copying_.) In the fourth place — `utarray_inserta` — the source object _is_ controlled by utarray, and so if you lack a `copy` ctor here you _will_ get double-free bugs. Still, maybe we intend to support users who are just really careful never to call `utarray_inserta`.

But, either way, it will be _useful in this test_ to provide a `copy` ctor, because otherwise you lack any test coverage for the `icd.copy` codepath in `utarray_replace`. So, please add a `Data_copy` function here. E.g.
```
void Data_copy(void* vdst, void* vsrc) {
    Data *dst = (Data*)vdst;
    Data *src = (Data*)vsrc;
    dst->id = src->id;
    dst->dept = strdup(src->dept);
}

UT_icd data_icd = {sizeof(Data), Data_init, Data_copy, Data_deinit};
```

And then remember to call `Data_deinit(&d)` on line 43 (pairing up with the `Data_init(&d)` on line 38).

Quuxplusone (changes requested):
Seems generally like a good idea.
Perhaps your documentation could also indicate that `utarray_replace(a,p,j)` has the same effect as `utarray_erase(a,j,1); utarray_insert(a,p,j)` but faster.

xunicatt:
Thanks a lot for the suggestions! I've updated the PR and addressed all the issues you pointed out. Let me know if there's anything else you'd like me to tweak. Appreciate the help!

Quuxplusone (approved):
LGTM now! It'll probably be a few more days before I get around to merging this, but I did a quick pass just now and I didn't see any problems with it.

Quuxplusone:
Squash-merged as af6e637f19c102167fb914b9ebcc171389270b48 — thanks @xunicatt!

PR #274, opened 2025-08-04
Update uthash.h to permit preallocation

sbergercnmc:
Placed HASH_INITIAL_NUM_BUCKETS and HASH_INITIAL_NUM_BUCKETS_LOG2 into a #ifndef block so as to permit these values to be optionally defined in other code before inclusion of uthash.h and thus permit preallocation of hashtable when a lot of keys are expected to be added.

### DIFF
--- a/src/uthash.h
+++ b/src/uthash.h
@@ -126,8 +126,14 @@ do {                                                                            
 #endif
 
 /* initial number of buckets */
+#ifndef HASH_INITIAL_NUM_BUCKETS
 #define HASH_INITIAL_NUM_BUCKETS 32U     /* initial number of buckets        */
+#endif
+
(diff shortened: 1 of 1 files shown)


Quuxplusone:
Seems plausible at first glance. Obvious questions:

- Have you used this in production? Does it help?
- Surely `HASH_BKT_CAPACITY_THRESH` should get the same treatment?
- Needs some test coverage.

sbergercnmc:
Doing this for my use case, where I am loading over 300,000,000 keys (uint64_t), my program's loading time goes from 178 seconds to 93 seconds...  Agreed that same idea can be applied to HASH_BKT_CAPACITY_THRESH, but I haven't needed this for my use case.

